PR #37, opened 2014-08-29
Enable searching for tickets by external_id

TheFrozenFire:
It is currently not possible to search for tickets by external_id, which is extremely important for API integrations (so you can retrieve tickets by their local record id). This patch adds in that capability.

### DIFF
--- a/src/Zendesk/API/Tickets.php
+++ b/src/Zendesk/API/Tickets.php
@@ -76,12 +76,16 @@ class Tickets extends ClientAbstract {
             $params['user_id'] = $this->client->users()->getLastId();
             $this->client->users()->setLastId(null);
         }
+        $queryParams = array();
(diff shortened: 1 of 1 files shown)


src/Zendesk/API/Tickets.php
```diff
@@ -76,12 +76,16 @@ public function findAll(array $params = array ()) {
             $params['user_id'] = $this->client->users()->getLastId();
             $this->client->users()->setLastId(null);
         }
+        $queryParams = array();
+        if($this->hasKeys($params, array('external_id'))) {
```
shanitang: Hi, thanks very much for your contribution, and sorry for the long delay. It look good, but I think it would be better to use isset($params['external_id']) instead of hasKey.

PR #41, opened 2014-09-18
Fixed bug where http_build_query encoding would remove parts of query strings

mortenson:
So trying this code:
```
$zendesk = new ZendeskAPI('acquia', $zendesk_creds['email']);
$zendesk->setAuth('token', $zendesk_creds['token']);
$query = 'status<solved+type:ticket';
$params = array(
  'query' => $query
);
$search = $zendesk->search($params);
```
throws this error:
```
Invalid search: Invalid status type: solved+type:ticket
```
It looks like the http_build_query call in HTTP::send is encoding the query in a way the API isn't expecting. 

I've added code to check if $json is an array in HTTP::send, and only in that case will http_build_query be called. This allows functions that call HTTP::send encode the GET params themselves, which I used to fix the issue with Search::performSearch.

It should be noted that the actual query I'm making is quite long, and this is the only way I've gotten it to work with the API.

Please let me know if anything needs to be changed in the PR.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -60,6 +60,8 @@ class Http {
             $json = new \stdClass();
         } else if ($contentType == 'application/json' && $method != 'GET' && $method != 'DELETE') {
             $json = json_encode($json);
+        } else if (is_array($json)) {
+          $json = http_build_query($json);
         }
 
         if ($method == 'POST') {
@@ -82,7 +84,7 @@ class Http {
             curl_setopt($curl, CURLOPT_POSTFIELDS, $json);
         } else {
             $curl = curl_init(
-                $url . ($json != (object)null ? (strpos($url, '?') === false ? '?' : '&') . http_build_query($json) : '')
+                $url . ($json != (object)null ? (strpos($url, '?') === false ? '?' : '&') . $json : '')
             );
             curl_setopt($curl, CURLOPT_CUSTOMREQUEST, ($method ? $method : 'GET'));
         }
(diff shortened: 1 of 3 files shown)


ggrossman:
@mortenson Thanks for your contribution and sorry for the long delay in getting it merged!

amcastror:
Guys, maybe I'm doing something wrong, but this brakes a search like "status:open". Let me try to explain:

If I try:
```php
    $query = 'status:open';
    $params = array(
      'query' => $query
    );
    $tickets = $client->search($params);
```
I get an incorrect result, because the final encoded url is:
```url
https://company.zendesk.com/api/v2/search.json?0=query%3Dstatus%3Aopen
```

The problem is the '0=' in the front. This is caused in [this line](https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Http.php#L69), because the variable $json is not really a json but a string like "query=status:open", so when converting it to a json you get this:
```
array(1) { [0]=> string(17) "query=status:open" }
```
Hence, the '0=' in front of the query.

A simple way to fix this is to revert [this method](https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Search.php#L27) to what it was before this PR. I realize that this raises the same issue than before, but that issue can be fixed by changing the "+" sign with a "," (coma).

Am I missing something here?? Thanks a lot!

ha-quidco:
Hi @amcastror,

could you have a look at the current master?
I've had similar issues with the search and think it might be fixed it with pull request #70.

As mentioned in the documentation, separate the search query parts by *space* for non-CLI requests instead of a *plus*.

Example: 
```
$query = 'type:ticket status:open';
    $params = array(
      'query' => $query
    );
    $tickets = $client->search($params);
```

Regards,
Hartmut

PR #61, opened 2015-01-13
Revise test files

shanitang:
Use BasicTest for all test class constructor including credentials setting and testing.
Change authentication settings to be environment variables, so those info can be set up in Travis CI as environment variables as well.
Revise most dependent test functions to have independent setup functions. 
Replace hard code objects' ids with dynamic created objects. 

@zendesk/archer


ggrossman:
:+1: let's get this merged!

vkmita:
:tada: :boom:

PR #63, opened 2015-01-15
Attachment filenames urlencoded

davesidious:
I added a call to urlencode() on the $params['name'] parameter to allow filenames with spaces etc. to be uploaded to Zendesk.


shanitang:
It seems there are some curl errors. Could you check on that again? Thank you!!

PR #66, opened 2015-01-20
Olaf/apps tests

ocke:
I wanted to add Apps test (/apps.json)
It didn't work, so I had to bugfix the http::send method
Then I wanted to write a test for Http
But it used procedural curl functions, so I had to write a CurlRequest class
Resulting in stubbing that class for testing

Long story short, I had to rewrite a whole bunch of things, but I tried to keep it to a minimum.

Image of fixing


ggrossman:
Hi @ocke, thanks for your contribution, looks like a lot of work and an interesting new direction for the test suite. We'll review it shortly!

ocke:
Hey @ggrossman,

One of the things would be to split live (functional) tests from actual unit tests. That's also the reason currently on github everything fails.

I would suggest, move Tests/ -> LiveTest/ and UnitTests/ -> Tests/. That way we can stick all the unit tests in the Tests/ directory and only when somebody wants to do an actual live tests, they will have to run 'phpunit tests/Zendesk/API/LiveTests/'. But by default it won't run those. Anyway, let me know what you guys think.

ggrossman:
Yes, we would like to redo the test suite along these lines. This is the way our [Ruby client library](https://github.com/zendesk/zendesk_api_client_rb) is structured, and only the unit tests run as part of the Travis-CI integration.

ggrossman:
And thanks for your effort at dealing with the test suite as it currently stands and helping to improve it :)

ocke:
Guess these are the first tests that pass travis :+1: 
I removed a lot of whitespace and closing php brackets
Added BasicTest :-)

ggrossman:
@ocke, this looks like great stuff, thank you!
We'll review it ASAP.
Here at Zendesk we're excited to see some new energy around this library!

ocke:
I drive the Zendesk Ruby team nuts with all my pull requests, so I'm switching to the Zendesk PHP team now ;-).

shanitang:
@ocke 
Thanks for all your work! It will be a great template for the following test file :)

PR #67, opened 2015-01-27
Add access to SatisfactionRatings API. Add satisfaction filter params.

chrisminett:
New method `Client::satisfactionRatings()` provides access to the SatisfactionRatings API class, which wasn't publicly available.

`SatisfactionRatings::findAll()` now accepts the [documented filter params](https://developer.zendesk.com/rest_api/docs/core/satisfaction_ratings#filter).

### DIFF
--- a/src/Zendesk/API/Client.php
+++ b/src/Zendesk/API/Client.php
@@ -39,6 +39,7 @@ namespace Zendesk\API;
  * @method OAuthTokens oauthTokens()
  * @method OrganizationFields organizationFields()
  * @method Organizations organizations()
+ * @method SatisfactionRatings satisfactionRatings()
  * @method SharingAgreements sharingAgreements()
  * @method SuspendedTickets suspendedTickets()
  * @method Tags tags()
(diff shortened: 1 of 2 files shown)


src/Zendesk/API/Client.php
```diff
@@ -183,6 +184,10 @@ class Client {
      */
     protected $organizations;
     /**
+     * @var SatisfactionRatings
+     */
+    protected $satisfactionRatings;
```
shanitang: Hi Chris,
Since we have SatisfactionRatings declared in class Tickets already, do you need SatisfactionRatings here for some purposes?
Thanks!!

chrisminett: Although it's declared in `Tickets`, it's not used or made available, which is why I've made the class available in the main `Client` class like the other services.

My particular use case is not related to any one ticket, but instead getting all satisfaction ratings with a comment, so they can be displayed on a office wallboard.

shanitang:
Yes, that makes sense! Thanks for your work :)

PR #68, opened 2015-02-02
Added Unit Test for Groups with Mocking

buheryfi:
Added Unit Test for Groups with a Mock Object so the test is not dependent on live HTTP Requests


ggrossman:
Cool! Any idea why Travis is failing?

buheryfi:
I'm guessing the failure is because the items on Line 13 and Line 17 are dependent on the BasicTest, which in turn requires credentials in phpunit.xml.dist.  So perhaps the Basic Test needs to be modified in the UnitTests folder to not require those credentials?

shanitang:
@buheryfi  Thank you!!

PR #72, opened 2015-02-17
Adding back support for methods that send parameter filename, example At...

lookbadgers:
Adding back support for methods that send parameter filename, example Attachments.

Latest version of Http::send does not support filename, Attachments requires support for filename or uploads do not work

Http::send($this->client, $endPoint, array('filename' => $params['file']), 'POST', (isset($params['type']) ? $params['type'] : 'application/binary'));

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -61,7 +61,7 @@ class Http {
 
         if ($method === 'POST') {
             $curl->setopt(CURLOPT_POST, true);
-
+            
         } else if ($method === 'PUT') {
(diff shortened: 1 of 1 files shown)


maximeprades:
cc @zendesk/archer

shanitang:
@lookbadgers Thanks you!

dpawluk:
did this not break any tests? @shanitang @lookbadgers I can't seem to get file uploads to work now.

dpawluk:
I believe `$params['file']` is the file uri and not the filename, so why are we assigning it to filename?

PR #73, opened 2015-02-23
added GroupMemberships Unit Tests with Mock

buheryfi:
added GroupMemberships Unit Tests with Mock


buheryfi:
I believe this failed because the withConsecutive() function was added in PHPUnit 4.1.0 and TravisCI is using 3.7.38.  See release note https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-4.1.0

Might it be time to update Travis to use a more current version of PHPUnit?

PR #75, opened 2015-02-24
added Unit Tests with Mock, Updated Organizations Endpoint

buheryfi:
added Unit Tests with Mock - Group Memberships and Organizations
changed Organizations->CreateMany() to use Plural Object rather than Single Object because the method was failing.


shanitang:
@buheryfi  Thanks!!

PR #76, opened 2015-02-25
Fix bugs for uploading

shanitang:
Fix bugs for uploading with `filename` and `body`

@zendesk/archer

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -87,8 +87,14 @@ class Http {
                 $size = filesize($filename);
                 $fileData = fread($file, $size);
                 $json = $fileData;
-                $contentType = '';
-            }            
+                $curl->setopt(CURLOPT_INFILE, $file);
+                $curl->setopt(CURLOPT_INFILESIZE, $size);
+            }
+            else if (isset($json['body'])){
+                $curl->setopt(CURLOPT_INFILESIZE, strlen($json['body']));
+                $json = $json['body'];
+            }
+
             $httpHeader[] = 'Content-Type: '.$contentType;
         } else {
             $contentType = '';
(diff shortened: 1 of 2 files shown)


kssmith:
Any update on this getting merged? We're running into problems on our production.

src/Zendesk/API/Http.php
```diff
@@ -87,8 +87,11 @@ public static function send(Client $client, $endPoint, $json = array(), $method
                 $size = filesize($filename);
                 $fileData = fread($file, $size);
                 $json = $fileData;
-                $contentType = '';
```
mikealmond: Need to tell cURL the file size as well

```php
curl_setopt($curl, CURLOPT_INFILESIZE, $size);
```

src/Zendesk/API/Http.php
```diff
@@ -87,8 +87,11 @@ public static function send(Client $client, $endPoint, $json = array(), $method
...
-            }            
+            }
+            else if (isset($json['body'])){
+                $json = $json['body'];
```
mikealmond: Need to set the size here

```php
curl_setopt($curl, CURLOPT_INFILESIZE, strlen($json['body']));
```

src/Zendesk/API/Http.php
```diff
@@ -87,8 +87,11 @@ public static function send(Client $client, $endPoint, $json = array(), $method
...
+            }
+
             $httpHeader[] = 'Content-Type: '.$contentType;
         } else {
             $contentType = '';
```
mikealmond: Does `$json` need to be reset with `$json['uploaded_data']`? It was before the change to this file
https://github.com/zendesk/zendesk_api_client_php/blob/552a5cad477b53f7ac7217c41c94529da8f28f2b/src/Zendesk/API/Http.php

src/Zendesk/API/Http.php
shanitang: @mikealmond 
Yes, you're right. Thanks!
But do you know why it still works without those set up in this case?

PR #77, opened 2015-02-27
Fix attachment upload

kssmith:
Content type was being reset on attachment upload where file type was passed to `Http::send()`
This causes the curl request response code of 500 if trying to send a file.


kssmith:
Closing in favour of #76

Issue #58, opened 2014-12-13
Is Incremental Ticket API supported?


shehi:
Title says it all: is Incremental Ticket API supported by this library? I checked the code thoroughly and couldn't find any sign of it in there. Thanks!

@see https://developer.zendesk.com/rest_api/docs/core/ticket_export

davidstanley01:
That link above is going to a 404 page, now.  Is the endpoint still good? Or should we be hitting the `incremental/tickets.json` endpoint and the feature in #80  instead?

PR #80, opened 2015-03-31
add incremental export feature

bvarent:
Adds support for [incremental exports](https://developer.zendesk.com/rest_api/docs/core/incremental_export) for tickets, users and organizations. Which is something different from the current /export implemented for Tickets.

also see issue #58

### DIFF
--- a/tests/Zendesk/API/LiveTests/TicketsTest.php
+++ b/tests/Zendesk/API/LiveTests/TicketsTest.php
@@ -170,6 +170,13 @@ class TicketsTest extends BasicTest {
         $this->assertEquals($this->client->getDebug()->lastResponseCode, '200', 'Does not return HTTP code 200');
     }
 
+    public function testIncremental() {
+        $tickets = $this->client->tickets()->export(array('start_time' => '1332034771'));
+        $this->assertEquals(is_object($tickets), true, 'Should return an object');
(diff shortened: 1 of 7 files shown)


tests/Zendesk/API/LiveTests/TicketsTest.php
```diff
@@ -170,6 +170,13 @@ public function testExport() {
         $this->assertEquals($this->client->getDebug()->lastResponseCode, '200', 'Does not return HTTP code 200');
     }
 
+    public function testIncremental() {
+        $tickets = $this->client->tickets()->export(array('start_time' => '1332034771'));
```
shanitang: Is this test for `incremental ` or `export` ?

shanitang:
@bvarent  Thanks for adding this!! Could you please add unit tests as well, because we only run unit tests for Travis CI?

buheryfi:
@bvarent, I've noticed that you've added the Incremental endpoints as a method to the class for users, organizations, and tickets, rather than creating a separate class for Incremental.  Our documentation shows it as a separate class and our client wrapper also implements separate classes in general, so it would be best to add a separate Incremental Class for these endpoints rather than including a method in existing classes.  Thoughts?

bvarent:
@buheryfi: Agreed. I think I can find time somewhere this week to turn it into a separate endpoint class.

atroche:
@bvarent let us know if you're still keen on doing this. I'll close the PR for now.

PR #81, opened 2015-04-01
Add support for the Push Notifications Devices destroy_many call.

andreionut:
/cc @ggrossman @shanitang 

This is for an upcoming API call: https://developer.zendesk.com/rest_api/docs/core/push_notification_devices

### DIFF
--- a/src/Zendesk/API/PushNotificationDevices.php
+++ b/src/Zendesk/API/PushNotificationDevices.php
@@ -17,7 +17,7 @@ class PushNotificationDevices extends ClientAbstract {
     }
 
     /**
-     * Delete a ticket or series of tickets
+     * Unregister the mobile devices that are receiving push notifications
      *
      * @param array $devices
      *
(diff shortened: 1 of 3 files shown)


shanitang:
@andreionut Thanks for this! Could you add a unit test as well?

andreionut:
bump

src/Zendesk/API/PushNotificationDevices.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @param Client $client
+     */
+    public function __construct(Client $client) {
+        parent::__construct($client);
+    }
+
+    /**
+     * Delete a ticket or series of tickets
```
shanitang: This should be "Unregister the mobile devices that are receiving push notifications" ?

andreionut: Yes, thanks for pointing this out.

andreionut:
We went live with the API endpoint today. If there aren't any objections can I get a +1 so I can merge?

shanitang:
:+1:

PR #82, opened 2015-04-08
Adds ability to use Debug class as a string

jurajseffer:
This is useful for logging. Conversion happens for lastResponseError if needed since all the other ones will be strings as option is passed in during curl_getinfo() calls.


shanitang:
Hi @jurajseffer  Thanks for your work. I'll go ahead and merge it.

PR #84, opened 2015-04-23
Add method for `users/show_many.json?external_ids=...`

mtibben:
The current `find` method does not allow using the `external_ids` param as documented at https://developer.zendesk.com/rest_api/docs/core/users#show-many-users

This adds a new method specifically for the `show_many` endpoint.

### DIFF
--- a/tests/Zendesk/API/LiveTests/UsersTest.php
+++ b/tests/Zendesk/API/LiveTests/UsersTest.php
@@ -17,7 +17,7 @@ class UsersTest extends BasicTest {
         parent::authTokenTest();
     }
 
-    protected $id, $id_s, $ticket_id, $number;
+    protected $id, $id_s, $ticket_id, $number, $external_id, $external_id_s;
 
     public function setUp() {
         $this->number = strval(time());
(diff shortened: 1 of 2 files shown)


mtibben:
Ugh, just realised I can query with `search` also.

Gotta be honest with you, the quality of this SDK is questionable. I find the REST API so far to be good, but this SDK really needs some work

tests/Zendesk/API/LiveTests/UsersTest.php
```diff
@@ -81,6 +85,20 @@ public function testFindMultiple() {
         $this->assertEquals(is_object($users->users[0]), true, 'Should return an object as first "users" array element');
     }
 
+    public function testShowManyUsingIds() {
+        $users = $this->client->users()->showMany(array('ids' => array($this->id, $this->id_s)));
```
atroche: hey @mtibben, is there a way you could DRY these two tests up? Seems like the only thing that changes between the two functions is the string `'id'` or `'external_id'`.

atroche:
:+1: looks solid

shanitang:
@mtibben Thanks for this !! :tada:

PR #90, opened 2015-05-31
Reformat code to match PSR-2

atroche:
PHP has an accepted community style guide, so I've reformatted our source files (using PhpStorm) to match it. Seems to be mostly about spacing between operators and braces going on the next line.

/cc @miogalang @joseconsador 

### References
 - Jira link: 

### Risks
 - None


miogalang:
:+1: yay PHPStorm

PR #92, opened 2015-06-02
Print debug info on any failure

atroche:
Not ideal, but a quick win: whenever we get an error response from the API, print debug information and throw an exception. Previously every single time we made a call we were doing the error checking and not providing any information, now we can do it in one place.

Todo: remove manual error checking from every API call… (like I've done with AuditLogs)

cc @miogalang @joseconsador

### DIFF
--- a/src/Zendesk/API/AuditLogs.php
+++ b/src/Zendesk/API/AuditLogs.php
@@ -56,9 +56,6 @@ class AuditLogs extends ClientAbstract
         }
         $endPoint = Http::prepare('audit_logs/' . $params['id'] . '.json');
         $response = Http::send($this->client, $endPoint);
-        if ((!is_object($response)) || ($this->client->getDebug()->lastResponseCode != 200)) {
-            throw new ResponseException(__METHOD__);
-        }
         $this->client->setSideload(null);
 
         return $response;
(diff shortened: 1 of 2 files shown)


miogalang:
:+1:

joseconsador:
:+1:

bilosnic:
what a shitty idea to print the debug message without any usage of display_errors or something else.
https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Http.php#L149

zendesk has heavy problems at the moment and is throwing 404 from time to time, an exception is supreme. but why would someone print debug to the customer if response code is > 400. bad bad bad idea :-1:

atroche:
@janson-git @bilosnic agreed, this was a temporary fix that hasn't made its way into a release yet and probably won't. are you sure you want to be requiring `dev-master` for whatever library you're using? I'd recommend using the latest release.

it's awesome that you took the time to help out — any chance you'd make a PR improving this, @bilosnic ?

it might not be worth it for you, because we're working on [v2 of the client](https://github.com/zendesk/zendesk_api_client_php/pull/99), which is using Guzzle for exceptions. but if you feel like making the improvement to that branch, that'd be amazing.

PR #98, opened 2015-06-03
add ViewsTest, update UsersTest for response code option

joseconsador:
Add ViewsTest, update UsersTest for response code option

### DIFF
--- a/tests/Zendesk/API/LiveTests/ViewsTest.php
+++ b/tests/Zendesk/API/LiveTests/ViewsTest.php
@@ -9,21 +9,12 @@ use Zendesk\API\Client;
  */
 class ViewsTest extends BasicTest
 {
+    protected $id = 12345;
 
-    public function testCredentials()
+    public function testCreate()
     {
-        parent::credentialsTest();
-    }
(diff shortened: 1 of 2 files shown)


miogalang:
:+1:

tests/Zendesk/API/LiveTests/ViewsTest.php
```diff
@@ -52,50 +43,53 @@ public function setUp()
...
             'Should return an object containing an array called "views"');
         $this->assertGreaterThan(0, $views->views[0]->id, 'Returns a non-numeric id for views[0]');
-        $this->assertEquals($this->client->getDebug()->lastResponseCode, '200', 'Does not return HTTP code 200');
     }
 
     public function testCompact()
     {
+        $this->mockApiCall('GET', '/views/compact.json?', array('views' => array(array('id' => $this->id))));
         $views = $this->client->views()->findAll(array('compact' => true));
         $this->assertEquals(is_object($views), true, 'Should return an object');
```
atroche: i don't think we need to check this every time — after all, if it doesn't return an object then the test is going to fail anyway, right?

atroche:
:+1:

PR #99, opened 2015-06-09
V2 - Rewrite of the PHP API Client

miogalang:
This branch is a complete rewrite of the PHP API Client with the following goals in mind:
* Make the tests run faster. Which means mocking the API calls, reducing the number of tests and making them easier to write.
* Make the code as DRY (Do not Repeat Yourself) as possible and improve maintainability.
* Improve the way we handle chaining.
* Better handling of routes.
* Handle errors well.

Basically we just wanted PHP developers using the client to actually **like** using the client.

Please comment at will. The more feedback, the better.

For our thoughts check out [our blog post](https://developer.zendesk.com/blog/chat-sdk-going-public)

/cc @joseconsador @jwswj @mmolina @atroche @buheryfi @shanitang


atroche:
@ocke as promised ;)

@joseconsador @miogalang can we get the travis build passing? are the tests running on your machines?

miogalang:
@atroche It seems travis is running the wrong tests. 

Here is a fix I came up with to fix it. https://github.com/zendesk/zendesk_api_client_php/pull/100

jwswj:
We're now at a point where we're planning to merge this new version of the API client into be the master branch. The current version of the API client will be moved to a v1 branch and maintained for now.

miogalang:
Good news. We are back to a merge-able state onto master. I would agree with @jwswj that this is ready for merging

PR #101, opened 2015-06-10
Add Ticket Fields resource

miogalang:
Add the Ticket Fields resource to the resources. How short it is is a good sign that we are DRYing this up.

### References
 - Jira link: 

### Risks
 - The ticket fields might not be accessible (low)


joseconsador:
:+1:

tests/Zendesk/API/MockTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+namespace Zendesk\API\MockTests;
+
+use Zendesk\API\Resources\TicketFields;
+
+/**
+ * Ticket Audits test class
```
mmolina: Ticket Fields test class :) Copy/Paste fail!! :stuck_out_tongue:

miogalang:
@mmolina :haircut:

mmolina:
:+1:

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+    public function testCreate()
+    {
+        $postFields = [
+          'type'  => 'text',
+          'title' => 'Age'
+        ];
+        $this->mockApiCall(
+          'POST',
+          'ticket_fields.json',
+          [ 'ticket_field' => [ 'id' => $this->id, 'type' => $postFields['type'], 'title' => $postFields['title'] ] ],
```
jwswj: Can we put these on a new line?

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+          [ 'bodyParams' => [ TicketFields::OBJ_NAME => $postFields ], 'statusCode' => '201' ]
+        );
+
+        $field = $this->client->ticketFields()->create( $postFields );
+        $this->httpMock->verify();
+        $this->assertEquals( is_object( $field ), true, 'Should return an object' );
```
jwswj: Is this assertion really needed? The next one also covers it

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->assertEquals( is_object( $field->ticket_field ), true, 'Should return an object called "ticket_field"' );
+        $this->assertGreaterThan( 0, $field->ticket_field->id, 'Returns a non-numeric id for ticket_field' );
```
jwswj: should this be `Returns a numeric id for the ticket_field`

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public function testAll()
+    {
+        $this->mockApiCall(
+          'GET',
+          'ticket_fields.json',
+          [ 'ticket_fields' => [ [ 'id' => 1 ] ] ]
+        );
+        $fields = $this->client->ticketFields()->findAll();
```
jwswj: noob question, but why is it findAll() vs list()?

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+        $this->assertEquals( $field->ticket_field->type, 'text', 'Type of test ticket field does not match' );
+        $this->assertEquals( $field->ticket_field->title, 'Age', 'Title of test ticket field does not match' );
+    }
+
+    public function testAll()
+    {
+        $this->mockApiCall(
+          'GET',
+          'ticket_fields.json',
+          [ 'ticket_fields' => [ [ 'id' => 1 ] ] ]
```
jwswj: Test data here should probably include more than 1 ticket field

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+        $fields = $this->client->ticketField( $this->id )->find();
+        $this->assertEquals( is_object( $fields ), true, 'Should return an object' );
+        $this->assertEquals( is_object( $fields->ticket_field ), true,
+          'Should return an object called "ticket_field"' );
+        $this->assertEquals( $this->id, $fields->ticket_field->id, 'Returns an incorrect id in ticket field object' );
+    }
+
+    public function testUpdate()
+    {
+        $bodyParams = [ 'title' => 'Another value' ];
```
jwswj: $bodyParams here, $postFields earlier. would be good to have consistency in the naming

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
miogalang: That is actually the error message to be displayed if the assertion fails. So if it isn't greater than 0 then it would say that a non-numeric id was returned.

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $this->assertEquals( is_object( $field ), true, 'Should return an object' );
+        $this->assertEquals( is_object( $field->ticket_field ), true, 'Should return an object called "ticket_field"' );
+        $this->assertGreaterThan( 0, $field->ticket_field->id, 'Returns a non-numeric id for ticket_field' );
+        $this->assertEquals( $field->ticket_field->type, 'text', 'Type of test ticket field does not match' );
+        $this->assertEquals( $field->ticket_field->title, 'Another value',
+          'Title of test ticket field does not match' );
+    }
+
+    public function testDelete()
```
jwswj: no assertion for this test?

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
miogalang: Honestly, this is because this was how it was done by v1 of the client. We can change it to anything we want.

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
miogalang: This is what we set the endpoint to return. so instead of making the API call we just return this. so in all honesty it doesn't even matter if this is correct or not. We just prefer to set a few fields.

tests/Zendesk/API/LiveTests/TicketFieldsTest.php
miogalang: Asserts are made by the mockAPICall(). It asserts a few things, The URL is correct, The method used for the call, and that the endpoint was called. That pretty much covers the asserts we need for delete.

In other cases you can test the post data and query data as well.

miogalang:
:haircut: @joseconsador @jwswj As discussed, removed the test and only need to test that the resource name was set correctly. The test for the base class should take care of the other tests.

jwswj:
:+1:

atroche:
loving the thoroughness of these reviews ;)

atroche:
:+1:

PR #102, opened 2015-06-11
Change the way we instantiate resources. 

miogalang:
Create new instances of each resource when calling them. 
`$client->tickets()->comments()` would instantiate new classes for `Tickets` and `Comments`. 
This means we do not have to worry about leftover state variables and is better design overall.

Additionally we want to add consistency and always use plural both for the resources class names and when instantiating them.

### References
 - Jira link: 

### Risks
 - We might not be able to chain properly (medium)

### DIFF
--- a/src/Zendesk/API/HttpClient.php
+++ b/src/Zendesk/API/HttpClient.php
@@ -7,6 +7,9 @@ namespace Zendesk\API;
  * spl_autoload_register(function($c){@include 'src/'.preg_replace('#\\\|_(?!.+\\\)#','/',$c).'.php';});
  */
 
+use Zendesk\API\Resources\Tickets;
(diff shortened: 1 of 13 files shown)


src/Zendesk/API/HttpClient.php
```diff
@@ -16,41 +19,8 @@
  * @method Debug debug()
  * @method Tickets ticket()
  * @method Tickets tickets()
- * @method TicketFields ticketFields()
```
miogalang: Removed all the stuff unavailable in V2. Add these as you add resources.

joseconsador:
:+1:

atroche:
:+1:

jwswj:
:+1:

PR #107, opened 2015-06-15
Remove var_dump(), pass Guzzle exception to our own exception class

joseconsador:
Remove var_dump() when an exception is thrown.
Wrap Guzzle exception to our own exception class.

### Risks
 - Medium - Existing classes for v2 that rely on exception handling may work differently

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -2,27 +2,30 @@
 
 namespace Zendesk\API;
 
+use Zendesk\API\Exceptions\ResponseException;
+
 /**
  * HTTP functions via curl
+ *
  * @package Zendesk\API
  */
 class Http
 {
     public static $curl;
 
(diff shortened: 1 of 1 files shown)


src/Zendesk/API/Http.php
```diff
@@ -90,31 +93,29 @@ public static function send_with_options(
...
-            null
-        );
+            $client->setDebug(
+                $response->getHeaders(),
+                $responseCode,
+                10,
+                null
+            );
+        } catch (\GuzzleHttp\Exception\ClientException $e) {
+            throw new ResponseException($endPoint, null, null, $e);
```
miogalang: Maybe log the contents of the client exception so we can find the error somewhere

miogalang:
:+1:

jwswj:
:+1:

PR #110, opened 2015-06-20
Added support for creating many tickets

jamesharv:
Just a simple addition to support the [tickets/create_many.json](https://developer.zendesk.com/rest_api/docs/core/tickets#creating-many-tickets) endpoint.

The new `Tickets::createMany()` method is based on the existing `Tickets::create()` method. The main difference being that it doesn't support attachments.


miogalang:
:+1:

joseconsador:
:+1:

PR #112, opened 2015-06-23
Update guzzle and mocking to use new guzzle

miogalang:
Updated guzzle to ~v6. This uses the PSR7 standard.
Additionally used the built in mocking mechanism of guzzle instead of an external library
Updating all the existing tests to use the guzzle mokcing mechanism

### References
 - Jira link: 

### Risks
 - Requests might not be sent properly (medium)


joseconsador:
:+1:

jwswj:
:+1:

PR #115, opened 2015-06-25
[MI-105] Add support for authentication with the api client

miogalang:
Add support for basic authentication and oauth token authentication in the client
Add live tests for checking authentication

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-105

### Risks
 - We might not be able to authenticate properly with the client (medium)

### DIFF
--- a/tests/Zendesk/API/LiveTests/BasicTest.php
+++ b/tests/Zendesk/API/LiveTests/BasicTest.php
@@ -1,0 +1,72 @@
+<?php
+
+namespace Zendesk\API\LiveTests;
+
+use Zendesk\API\HttpClient;
+
+/**
+ * Basic test class
+ */
+abstract class BasicTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var HttpClient
+     */
+    protected $client;
+    /**
+     * @var string
+     */
+    protected $subdomain;
+    /**
+     * @var string
+     */
+    protected $password;
+    /**
+     * @var string
+     */
+    protected $token;
+    /**
+     * @var string
+     */
+    protected $oAuthToken;
+    /**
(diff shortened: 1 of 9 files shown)


joseconsador:
:+1:

miogalang:
:+1: to @joseconsador 's commit

tests/Zendesk/API/LiveTests/BasicTest.php
```diff
@@ -0,0 +1,72 @@
...
+    protected $mockedTransactionsContainer = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct()
+    {
+        $this->subdomain  = getenv('SUBDOMAIN');
+        $this->username   = getenv('USERNAME');
+        $this->password   = getenv('PASSWORD');
```
jwswj: Do we support password auth? Looks like just token.

jwswj:
Nice :+1:

tests/Zendesk/API/LiveTests/BasicTest.php
miogalang: Do we want to support password auth? At the moment we don't but it is easy to add.

miogalang: Yup only token for now. Do we want to add password auth? it should be pretty straightforward. But removing this for now

jwswj: I don't think we should. There is nothing worse than having clear text passwords in code. If it comes up we should be moving people to token based authentication for API clients.

miogalang: I agree 100%

PR #119, opened 2015-06-29
[MI-144] Allow client to use the Triggers resource

miogalang:
Add the triggers resource and allow usage of the trigger related endpoints

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-144

### Risks
 - We might not be able to call the trigger endpoints correctly (low)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ php:
   - 5.5
   # - 5.4
   # - hhvm
-before_script: composer self-update && composer install
+before_script:
+  - travis_retry composer self-update && composer install
 script:
   - vendor/bin/phpunit --testsuite "Zendesk API Unit Test Suites"
   - vendor/bin/phpcs --extensions=php --standard=PSR2 --report=summary -np src/ tests/
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,7 @@ php:
   - 5.5
   # - 5.4
   # - hhvm
-before_script:
-  - travis_retry composer self-update && composer install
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,7 +4,8 @@ php:
   - 5.5
   # - 5.4
   # - hhvm
-before_script: composer self-update && composer install
+before_script:
+  - travis_retry composer self-update && composer install
```
jwswj: What does `travis_retry` do? Why is it needed for this PR?

src/Zendesk/API/HttpClient.php
```diff
@@ -143,11 +148,12 @@ public function __construct(
...
-            'users'   => Users::class,
-            'views'   => Views::class,
-            'tags'    => Tags::class,
-            'macros'  => Macros::class,
+            'tickets'  => Tickets::class,
+            'users'    => Users::class,
+            'views'    => Views::class,
+            'tags'     => Tags::class,
+            'macros'   => Macros::class,
+            'triggers' => Triggers::class,
```
jwswj: This is why aligning equal signs sucks.

jwswj:
:+1:

.travis.yml
miogalang: We keep on getting failures of travis builds because of network failures when installing composer dependencies. We then manually retry the build. This will retry the command 3x before failing. 

I will pull this out into a separate PR

joseconsador:
:+1:

PR #120, opened 2015-06-29
[MI-120] Allows the client to use the Groups resource

joseconsador:
Adds the Group resource and unit test

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-120

### Risks
 - None

### DIFF
--- a/src/Zendesk/API/HttpClient.php
+++ b/src/Zendesk/API/HttpClient.php
@@ -9,6 +9,7 @@ namespace Zendesk\API;
 
 use Zendesk\API\Exceptions\AuthException;
 use Zendesk\API\Resources\Automations;
+use Zendesk\API\Resources\Groups;
 use Zendesk\API\Resources\Macros;
 use Zendesk\API\Resources\Tags;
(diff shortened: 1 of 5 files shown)


src/Zendesk/API/HttpClient.php
```diff
@@ -126,10 +128,10 @@ public function __construct(
         }
 
         $this->subdomain = $subdomain;
-        $this->username  = $username;
-        $this->hostname  = $hostname;
-        $this->scheme    = $scheme;
-        $this->port      = $port;
```
jwswj: I feel like this indentation was changed recently? Which one are we using?

jwswj:
:+1:

joseconsador:
:+1:

src/Zendesk/API/HttpClient.php
joseconsador: We had different spacing schemes set on PHPStorm. I've now matched it to align the equal sign.

miogalang:
:+1:

PR #122, opened 2015-06-29
[MI-130] Add user Identities Resource

miogalang:
Add the user identities resource

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-130

### Risks
 - We might not be able to call the user identity endpoint correctly (low)

### DIFF
--- a/src/Zendesk/API/Resources/UserIdentities.php
+++ b/src/Zendesk/API/Resources/UserIdentities.php
@@ -1,0 +1,206 @@
+<?php
+
+namespace Zendesk\API\Resources;
+
+use Zendesk\API\Exceptions\MissingParametersException;
+use Zendesk\API\Http;
+
+/**
+ * The UserIdentities class exposes fields on the user profile page
+ * @package Zendesk\API
+ */
+class UserIdentities extends ResourceAbstract
+{
+
+    const OBJ_NAME = 'identity';
+    const OBJ_NAME_PLURAL = 'identities';
+
+    protected $resourceName = 'identities';
+
+    protected function setUpRoutes()
+    {
+        $this->setRoutes([
+            'findAll'             => "users/{userId}/{$this->resourceName}.json",
+            'find'                => "users/{userId}/{$this->resourceName}/{id}.json",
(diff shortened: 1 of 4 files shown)


src/Zendesk/API/Resources/UserIdentities.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * This API method allows you to set an identity to primary.
+     */
+    public function makePrimary(array $params = [])
+    {
+        $this->addUserIdToRouteParams($params);
+
+        if (isset($params['id'])) {
```
jwswj: This block is used 3 times. Could it be extracted into a private method?

tests/Zendesk/API/UnitTests/UserIdentitiesTest.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+
+        $this->mockAPIResponses([
+            new Response(200, [], '')
+        ]);
+        $userId = 3124;
+
+        $postFields = [
+            'type'  => 'email',
+            'value' => 'devaris.brown@zendesk.com'
```
jwswj: can we use another example email rather than an ex-employee?

miogalang:
:haircut: @jwswj

joseconsador:
:+1:

src/Zendesk/API/Resources/UserIdentities.php
```diff
@@ -0,0 +1,187 @@
...
+     * @param string $callingMethod
+     * @param array $params
+     *
+     * @return array
+     * @throws MissingParametersException
+     * @throws \Exception
+     * @throws \Zendesk\API\Exceptions\ApiResponseException
+     * @throws \Zendesk\API\Exceptions\AuthException
+     */
+    private function makePutRequest($callingMethod, $params = [])
```
jwswj: In Ruby private functions go under public functions to make it easier to read all the public interfaces at the top of the file, followed by the internal methods. I think it would be nice to use the same approach here.

jwswj:
:+1:

PR #126, opened 2015-06-30
Add bulk traits

joseconsador:
Adds traits for bulk actions. This way special cases can be handled individually.
Adds another wrapper for http requests under HttpClient . This serves to reduce repetition in some methods that only require a change in endpoint, so we don't have to call the ResourceAbstract methods like create() or find() . Instead $this->client->post() / get() / delete() / put() can be used.

### References
 - Jira link: 

### Risks
 - High - query parameters may not get passed to the api request

### DIFF
--- a/src/Zendesk/API/BulkTraits/BulkCreateTrait.php
+++ b/src/Zendesk/API/BulkTraits/BulkCreateTrait.php
@@ -11,7 +11,7 @@ use Zendesk\API\Exceptions\RouteException;
 trait BulkCreateTrait
 {
     /**
-     * Create multiple new respirces
+     * Create multiple new resources
      *
      * @param array $params
      *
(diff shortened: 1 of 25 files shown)


jwswj:
Love this change. Is there a way you could directly test the traits being introduced as they are the heart and soul of this API client?

miogalang:
see: https://phpunit.de/manual/current/en/test-doubles.html

Specifically the part with the header: 'Mocking Traits and Abstract Classes'

We could potentially move over our test for the abstract class to this pattern.

joseconsador:
@miogalang Thanks for the tip!

joseconsador:
Since the traits are dependent on ResourceAbstract, placed the tests on ResourceTest so there's no need to create another dummy resource.

jwswj:
:+1:

miogalang:
:+1:
Nice. Now we move active to its own trait as well

PR #128, opened 2015-07-02
[MI-134] Allow the client to use the organizations resource

joseconsador:
[MI-134] Allow the client to use the organizations resource

### References
 - Jira link: 

### Risks
 - Medium - the client may not get the expected API response

### DIFF
--- a/src/Zendesk/API/Resources/Organizations.php
+++ b/src/Zendesk/API/Resources/Organizations.php
@@ -24,11 +24,13 @@ class Organizations extends ResourceAbstract
     {
         parent::setUpRoutes();
 
-        $this->setRoutes([
-            'autocomplete' => $this->resourceName . '/autocomplete.json',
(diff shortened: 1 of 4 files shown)


miogalang:
:+1:

src/Zendesk/API/Resources/Organizations.php
```diff
@@ -0,0 +1,122 @@
...
+    use BulkUpdateTrait;
+    use BulkDeleteTrait;
+
+    protected function setUpRoutes()
+    {
+        parent::setUpRoutes();
+
+        $this->setRoutes([
+            'autocomplete' => $this->resourceName . '/autocomplete.json',
+            'related'      => "{$this->resourceName}/{id}/related.json",
```
jwswj: seems strange to use a different string concatenation approach - let's be consistent

jwswj:
:+1:

PR #130, opened 2015-07-02
[MI-121] Add job statuses resource and test

joseconsador:
Add job statuses resource and test

### References
 - Jira link: 

### Risks
  - Medium - the client may not get the expected API response


miogalang:
:+1:

src/Zendesk/API/Resources/JobStatuses.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Zendesk\API\Resources;
+
+use Zendesk\API\BulkTraits\BulkFindTrait;
+
+class JobStatuses extends ResourceAbstract
+{
+    const OBJ_NAME = 'job_status';
+    const OBJ_NAME_PLURAL = 'job_statuses';
```
atroche: is it just me or could `OBJ_NAME` and `OBJ_NAME_PLURAL` be inferred from the class name in most cases?

src/Zendesk/API/Resources/JobStatuses.php
```diff
@@ -0,0 +1,18 @@
...
+
+    use BulkFindTrait;
+
+    protected function setUpRoutes()
+    {
+        $this->setRoute('find', "{$this->resourceName}/{id}.json");
```
atroche: If there are lots of resources that have a find route that uses this exact same URL pattern, is there a way you could just be like “this resource has the following methods: [find]” and then it knows to do that already, without writing out a function? Like, just an array on the class or something?

atroche: then the tests below could just assert the contents of that array ;) rather than trying to call all the different standard functions

joseconsador: `$this->setRoute('find', "{$this->resourceName}/{id}.json");` is similar to what you said because it overrides the default available routes set in ResourceAbstract . If for example a method with a missing route is found, an exception is thrown.

:haircut:

src/Zendesk/API/Resources/JobStatuses.php
jwswj: Totally. Will add a story to investigate it.

jwswj:
:+1:

PR #129, opened 2015-07-02
[MI-107] Allow user to update profile photo using client

miogalang:
Allow the user to update user profile photo with the api client.
This separates the updating by specifying a file and uploading it, and updating from a remote url.

/cc @joseconsador @jwswj @mmolina @atroche 

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-107

### Risks
 - The user might not be able to update his profile image (Low)

PR #131, opened 2015-07-02
[MI-105] Move Auth operations to it's own utility class

miogalang:
Move the auth operations to utility class
Add unit test for auth

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-105

### Risks
 - We might not be able to make authenticated calls with the client (High)


src/Zendesk/API/Http.php
```diff
@@ -75,11 +76,11 @@ public static function sendWithOptions(
             $options
         );
 
-        $headers = [
+        $headers        = [
             'Accept'       => 'application/json',
             'Content-Type' => $options['contentType'],
         ];
-
+        $requestOptions = [];
```
miogalang: This is here because $requestOptions is used for the multipart request seen in #129

joseconsador:
:+1:

jwswj:
Great change :+1:

PR #132, opened 2015-07-02
[MI-135] Add organization fields resource

miogalang:
Add the organization fields resource so it can be used with the client

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-135

### Risks
 - We might not be able to call the organization resource endpoints using the client (low)


joseconsador:
:+1

jwswj:
:+1:

PR #139, opened 2015-07-06
[MI-108] Add activities resource

miogalang:
Add the activities resource so we can use the activities related endpoints using the client.

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-108

### Risks
 - Low. We might not be able to use the activities endpoints using the client


joseconsador:
:+1:

jwswj:
:+1:

PR #140, opened 2015-07-07
Move clearing of sideload to send utility method

miogalang:
Move clearing of sideload to every request made to reduce repeating code

### References
* JIRA:

### Risks
* Low. Sideloading might not work

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -130,6 +130,8 @@ class Http
             fclose($file);
         }
 
+        $client->setSideload(null);
+
         return json_decode($response->getBody()->getContents());
     }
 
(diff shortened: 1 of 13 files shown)


joseconsador:
:+1:

jwswj:
Are there any tests for this functionality?

joseconsador:
@jwswj There's ResourceTest::testSideLoad() that does some testing if sideloads are set.

jwswj:
What about an example test of sideloading for a ticket? This is key functionality that we need to make sure is stable for the client.

miogalang:
What I did here is add a test to check the clearing of the sideload after the query.

There is and existing test for tickets sideloading. `TicketsTest::testAllSideLoadedMethod` and `TicketsTest::testAllSideLoadedParameter` which test for using the `sideload()` method and passing it as a parameter to `findAll(['sideload' => 'something'])` respectively.

jwswj:
:+1:

PR #141, opened 2015-07-07
Move common crud functions to traits

miogalang:
Moving the common crud function to traits.
Re-arranging the traits into a traits folder.

This will allow us to include only the methods needed for the specific resource. 
e.g. AuditLogs only has `find` and `findAll` methods so we do not include `create`, etc. to that resource.

### References
* JIRA:

### Risks
* Medium. We might not be able to call the crud functionality correctly


joseconsador:
:+1 for mio's commits

miogalang:
:+1: for @joseconsador' s commits

src/Zendesk/API/Resources/AuditLogs.php
```diff
@@ -12,6 +14,12 @@ class AuditLogs extends ResourceAbstract
     const OBJ_NAME = 'audit_log';
     const OBJ_NAME_PLURAL = 'audit_logs';
 
+    use FindAll {
+        findAll as traitFindAll;
+    }
+
+    use Find;
```
jwswj: Can this go above FindAll so they are next to each other.

src/Zendesk/API/Resources/AuditLogs.php
```diff
@@ -12,6 +14,12 @@ class AuditLogs extends ResourceAbstract
     const OBJ_NAME = 'audit_log';
     const OBJ_NAME_PLURAL = 'audit_logs';
 
+    use FindAll {
+        findAll as traitFindAll;
```
jwswj: traitFindAll or findAllTrait?

src/Zendesk/API/Resources/Tickets.php
```diff
@@ -2,13 +2,14 @@
...
-use Zendesk\API\BulkTraits\BulkFindTrait;
-use Zendesk\API\BulkTraits\BulkUpdateTrait;
 use Zendesk\API\Exceptions\MissingParametersException;
 use Zendesk\API\Exceptions\ResponseException;
 use Zendesk\API\Http;
-use Zendesk\API\UtilityTraits\InstantiatorTrait;
+use Zendesk\API\Traits\Resource\Defaults;
+use Zendesk\API\Traits\Resource\DeleteMany;
+use Zendesk\API\Traits\Resource\FindMany;
+use Zendesk\API\Traits\Resource\UpdateMany;
```
jwswj: no `Zendesk\API\Traits\Resource\Update;`?

src/Zendesk/API/Resources/Views.php
```diff
@@ -13,6 +18,17 @@
  */
 class Views extends ResourceAbstract
 {
+    use FindAll {
+        findAll as traitFindall;
```
jwswj: traitFindAll

src/Zendesk/API/Traits/Resource/DefaultsMany.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Zendesk\API\Traits\Resource;
+
+/**
+ * This trait gives resources access to the default bulk CRUD methods.
+ *
+ * @package Zendesk\API\Traits\Resource
+ */
+trait DefaultsMany
```
jwswj: Might be overkill ?

tests/Zendesk/API/UnitTests/ViewsTest.php
```diff
@@ -215,22 +248,142 @@ public function testPreviewCount()
...
         );
 
         $this->assertEquals(is_object($preview), true, 'Should return an object');
         $this->assertEquals(is_object($preview->tickets), true, 'Should return an object called "tickets"');
     }
+
+    /**
+     * Test creating a view
+     */
+    public function testCreate()
```
jwswj: Why are these tests important for views? but not others?

tests/Zendesk/API/UnitTests/JobStatusesTest.php
```diff
@@ -1,14 +0,0 @@
-<?php
-
-namespace Zendesk\API\UnitTests;
-
-class JobStatusesTest extends BasicTest
-{
-    public function testFindMethodOnly()
```
jwswj: What test is this covered by?

tests/Zendesk/API/UnitTests/AttachmentsTest.php
```diff
@@ -85,19 +85,4 @@ public function testDeleteAttachment()
             ]
         );
     }
-
-    /**
-     * Test routes for find and delete are present
-     */
-    public function testRoutes()
```
jwswj: Isn't this still relevant?

src/Zendesk/API/Traits/Utility/ChainedParametersTrait.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Zendesk\API\UtilityTraits;
+namespace Zendesk\API\Traits\Utility;
```
jwswj: Great namespace update

src/Zendesk/API/Resources/AuditLogs.php
miogalang: This is a horrible name I agree.

The purpose of this is to differentiate between the findAll defined in `AuditLogs.php` and the findAll defined in the Trait `findAll.php`. 

findAllTrait sounds like a Trait but it is in fact a method.

Is there a better name you can think of? `genericFindAll` perhaps? or `generalFindAll`?

src/Zendesk/API/Resources/Tickets.php
joseconsador: `Zendesk\API\Traits\Resource\Defaults;` loads Find, FindAll, Create, Update, and Delete. We added `Defaults` so that most endpoints could just use that and not have to include everything.

jwswj: Ah - missed defaults.

src/Zendesk/API/Resources/Tickets.php
```diff
@@ -22,11 +23,15 @@
...
-    use BulkFindTrait;
-    use BulkUpdateTrait {
-        BulkUpdateTrait::updateMany as bulkUpdate;
+    use FindMany;
+    use UpdateMany {
+        UpdateMany::updateMany as bulkUpdate;
+    }
+    use DeleteMany;
+
+    use Defaults {
```
jwswj: I feel like defaults should always go first

src/Zendesk/API/Resources/Activities.php
```diff
@@ -11,14 +14,6 @@ class Activities extends ResourceAbstract
...
-     * Sets up the available routes for the resource.
-     */
-    protected function setUpRoutes()
-    {
-        $this->setRoutes([
-            'findAll' => "{$this->resourceName}.json",
-            'find'    => "{$this->resourceName}/{id}.json",
-        ]);
-    }
+    use FindAll;
```
jwswj: Sometimes these are above the const's other times they are below

src/Zendesk/API/Traits/Resource/DefaultsMany.php
jwswj: Only used twice. I'd just keep referencing them for now until a 3rd is added.

tests/Zendesk/API/UnitTests/ViewsTest.php
miogalang: These are actually not important here. I put them temporarily as I wanted to follow TDD in this case where I had a failing test and I had to make them pass one by one.

Let me remove them as they are actually tested in the DummyResource test.

src/Zendesk/API/Traits/Resource/DefaultsMany.php
joseconsador: Currently it's only used once. We can remove it if there are no other use-cases.

tests/Zendesk/API/UnitTests/JobStatusesTest.php
joseconsador: This used to test if the route was excluded properly, but since we're moving to just use inclusions, it's not relevant anymore. ResourceTest covers the Find and FindMany trait used in the JobStatuses endpoint.

src/Zendesk/API/Traits/Resource/DefaultsMany.php
joseconsador: Yup, twice including the DummyResource class.

tests/Zendesk/API/UnitTests/AttachmentsTest.php
joseconsador: Don't need this since it used to test if only the needed routes (methods) were available. But we're now including from traits and not dependent on route definition.

joseconsador:
:haircut:

src/Zendesk/API/Resources/Automations.php
```diff
@@ -9,15 +11,18 @@
  */
 class Automations extends ResourceAbstract
 {
+    use Defaults;
 
     const OBJ_NAME = 'automation';
     const OBJ_NAME_PLURAL = 'automations';
 
+    use Defaults;
```
jwswj: Included twice

src/Zendesk/API/Resources/Users.php
```diff
@@ -19,13 +21,20 @@
...
-
     const OBJ_NAME = 'user';
     const OBJ_NAME_PLURAL = 'users';
 
-    use BulkCreateTrait;
-    use BulkUpdateTrait;
+    use InstantiatorTrait;
+
+    use Defaults {
+        findAll as traitFindall;
```
jwswj: `traitFindAll`

jwswj:
After the 2 small changes :+1:

atroche:
like it. does this mean you can now just test those traits separately, and then for each resource just assert that it uses the right traits? would remove a lot of boilerplate tests, right?

:+1:

PR #154, opened 2015-07-13
[MI-127] Add the oauth clients resource

miogalang:
Add the oauth clients resource so we can call the endpoints using the client.

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-126

### Risks
* Low. We might not be able to call the OAuth clients endpoints using the client.


joseconsador:
:+1:

jwswj:
:+1:

PR #155, opened 2015-07-13
[MI-128] Custom Agent Roles

joseconsador:
Add Custom Agent Roles resource and test

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-128

### Risks
 - Low. We might not be able to call the Custom Roles endpoint using the client.


miogalang:
:+1: #beautifullySimple

jwswj:
:+1:

PR #156, opened 2015-07-14
[MI-131] Add brands resource

miogalang:
Add brands resource so we can use the brands endpoints using the client

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-131

### Risks
* Low. We might not be able to access the brands endpoints using the client


jwswj:
:+1:

joseconsador:
:+1:

PR #157, opened 2015-07-14
Use BasicTest::assertEndpointCalled() for most mocked tests

joseconsador:
Use a wrapper function to assert the API client functionality. This removes a lot of repetitive test code.

### References
 - Jira link: 

### Risks
 - Medium - the tests might not cover all cases

### DIFF
--- a/tests/Zendesk/API/UnitTests/AppInstallationsTest.php
+++ b/tests/Zendesk/API/UnitTests/AppInstallationsTest.php
@@ -2,7 +2,6 @@
 
 namespace Zendesk\API\UnitTests;
 
-use GuzzleHttp\Psr7\Response;
 use Zendesk\API\Resources\AppInstallations;
 
 class AppInstallationsTest extends BasicTest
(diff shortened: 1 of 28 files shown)


miogalang:
:+1: @joseconsador 's commit

jwswj:
This does reduce repetitive code, but it also makes it a little more complicated in my mind. I'm not 100% convinced.

miogalang:
Yes this adds a layer of complexity. But for me it is worth it for readability.

Basically all that is in the code is. This snippet of code would call this endpoint with these parameters. 

It would make the tests easier to understand and removes a lot of the unnecessary lines of code from the test cases.

jwswj:
@danielstjules I hear you've done a fair bit of php in the past. What are your thoughts?

jwswj:
:+1:

danielstjules:
@jwswj [Data providers](https://phpunit.de/manual/current/en/phpunit-book.html#writing-tests-for-phpunit.data-providers) might be helpful for some test classes, helping reduce some of the repetitive testing boilerplate.

danielstjules:
LGTM: :)

PR #161, opened 2015-07-15
[MI-141] Support Address

joseconsador:
Add support address resource and test

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-141

### Risks
 - Low. We might not be able to use the support address resources


jwswj:
:+1:

miogalang:
:+1:

PR #167, opened 2015-07-16
[MI-151] Add the SLA Policies resource

miogalang:
Add the SLA policies resource and test case

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-151

### Risks
* Low. We might not be able to call the sla policies endpoints using the client


jwswj:
:+1:

joseconsador:
:+1:

PR #184, opened 2015-08-07
Update readme and example to use the new auth style

jwswj:
When merging in the v2 branch we forgot to update the readme file to reference the new HTTPClient and auth which is used.

/cc @joseconsador @miogalang 

### Risks
 - The examples might actually work :)

### DIFF
--- a/index.php
+++ b/index.php
@@ -4,12 +4,11 @@ include("vendor/autoload.php");
 use Zendesk\API\HttpClient as ZendeskAPI;
 
 $subdomain = "subdomain";
-$username  = "username";
+$username  = "email@company.com";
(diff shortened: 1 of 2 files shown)


index.php
```diff
@@ -24,11 +23,3 @@
   'priority' => 'normal'
 ));
 print_r($newTicket);
-
-// Update multiple tickets
-$client->ticket(array(123, 456))->update(array(
```
jwswj: Deleting this example as you can't always be sure that a ticket with ID 123 and 456 exists.

miogalang:
:+1: can't believe we missed this

PR #191, opened 2015-08-21
Revert back to the 2.0.0-beta version

jwswj:
We moved to a 2.1.0-beta version in the master branch after merging some changes to the OAuth client recently. 2.0.0-beta1 is much more consistent as we aim to move to a stable 2.0.0 release in the coming months.

@joseconsador @miogalang 

## Risks
- Composer might get confused, but it [doesn't look](https://packagist.org/packages/zendesk/zendesk_api_client_php) like the latest version was pushed


miogalang:
:+1:

joseconsador:
:+1:

PR #192, opened 2015-08-25
Compatible with PHP 5.3

jonathan-s:
We've been running this on PHP 5.3 with no problems, so it's compatible with 5.3.0.


jonathan-s:
(saw guzzle requirements in later versions)

Issue #195, opened 2015-09-02
User Search via external_id no longer working


joho:
We have some code written against the v1 client to find users based on our external_id that no longer works (see below)
```php
$result = $this->client->users()->search(array(
    'external_id' => sprintf('user-%s', $externalId),
));
```

Comparing the v1 and v2 code, it appears now that you're whitelisting query params between `Http:: prepareQueryParams` (https://github.com/zendesk/zendesk_api_client_php/blob/700188225c77cf94caa87e7b833798cdf57905d5/src/Zendesk/API/Http.php#L27-L45) and the code in `Users->search` (https://github.com/zendesk/zendesk_api_client_php/blob/a1b69bc155632fcd7d3c045d8647d941c0171b21/src/Zendesk/API/Resources/Core/Users.php#L221-L222) and as such are stripping out the `external_id` from the query params.

For local dev env reasons I've had trouble getting things set up to verify, but I think that when there are no params sent to search it returns anything that matches (everything), or at least that's my hunch because all our tickets lodged by signed in users are getting attached to our oldest user.

Would appreciate if you could confirm/fix, or just close it off if I'm mistaken.

jwswj:
@joho legacy from v1, but it's a reasonable change so we'll have to spend a bit of time testing it. Can follow along #196 if you're interested or have any comments.

joho:
Any chance a specific fix can get released before a solve the world as I've got a fair volume of tickets being mis-assigned (or i could reject them going in for a nonsense match). If not I'll fork & hack it to temporarily working for the weekend.

jwswj:
@joho quick fix has been merged into master - https://github.com/zendesk/zendesk_api_client_php/pull/197

joho:
many many **many** thanks, i'll get that shipped asap.

PR #197, opened 2015-09-04
Whitelist external_id as an accepted query param

jwswj:
@miogalang @joseconsador @iandjx @joho 

Quick fix for https://github.com/zendesk/zendesk_api_client_php/issues/195 while we work through the bigger change.


miogalang:
:+1:

PR #198, opened 2015-09-11
change method name to tickets()

aj0y:
change method name to tickets() to conform to the list of valid sub resources


jwswj:
:+1:

PR #199, opened 2015-09-11
Add php7 into the build matrix

jwswj:
Add PHP7 into the build matrix, allowing it to fail for now.

@miogalang @joseconsador 

### Risks
- None


miogalang:
Wow eveything is actually passing on PHP 7. :+1:

Issue #201, opened 2015-09-21
PHP Fatal Error: Call to a member function getReasonPhrase() on a non-object


mtibben:
Using 04e31fba we're seeing the following error being thrown inside the Zendesk library:

```
vendor/zendesk/zendesk_api_client_php/src/Zendesk/API/Exceptions/ApiResponseException.php:22
Call to a member function getReasonPhrase() on a non-object
```

Given that it's a fatal error occurring inside the library we can't see the context in which this is occuring

joseconsador:
@mtibben Kindly provide us with more details so we can look further into the issue. What endpoints/methods were you trying to use? Are you able to make the sample (https://github.com/zendesk/zendesk_api_client_php/blob/master/samples/getTickets.php) work? Thanks

PR #202, opened 2015-09-22
Use guzzle's exception message

joseconsador:
Fixes the fatal error from #201 .
The response object from the ResponseException is NULL, which causes this error.
Instead of getting the fatal error, an exception in the following format will be received instead.

```
Zendesk\API\Exceptions\ApiResponseException: Error completing request [url] http://foo.com [http method] GET
```

Uses Guzzle's RequestException::create() method to generate the exceptions, below is a sample of a ClientException.

```
Zendesk\API\Exceptions\ApiResponseException: Client error response [url] https://z3njosemanila.zendesk.com/api/v2/tickets/99999999.json [http method] GET [status code] 404 [reason phrase] Not Found [details] {"error":"RecordNotFound","description":"Not found"}
```

Added a test for handling situations where the server returns no response or the connection is closed.

/cc @miogalang @jwswj @iandjx @pdeuter 

### References
 - Jira link: 

### Risks
 - Medium: Exceptions and cases not covered in ApiResponseException::__construct() may not be caught.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -85,8 +85,8 @@ class Http
             list ($request, $requestOptions) = $client->getAuth()->prepareRequest($request, $requestOptions);
             $response = $client->guzzle->send($request, $requestOptions);
         } catch (RequestException $e) {
-            $exception = RequestException::create($e->getRequest(), $e->getResponse());
-            throw new ApiResponseException($exception);
+            $requestException = RequestException::create($e->getRequest(), $e->getResponse());
+            throw new ApiResponseException($requestException);
         } finally {
             $client->setDebug(
                 $request->getHeaders(),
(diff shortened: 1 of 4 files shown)


src/Zendesk/API/Http.php
```diff
@@ -85,7 +85,8 @@ public static function send(
             list ($request, $requestOptions) = $client->getAuth()->prepareRequest($request, $requestOptions);
             $response = $client->guzzle->send($request, $requestOptions);
         } catch (RequestException $e) {
-            throw new ApiResponseException($e);
+            $exception = RequestException::create($e->getRequest(), $e->getResponse());
```
jwswj: $requestException might be a better variable name

jwswj:
looks good

src/Zendesk/API/Http.php
joseconsador: :haircut:

miogalang:
:+1:

jwswj:
:+1:

PR #213, opened 2015-10-20
Add request body to exception message

joseconsador:
Adds the request body to the exception message, in cases where no response is returned from the API call.

/cc @miogalang @jwswj @mmolina @iandjx @pdeuter

### References
 - Jira link: 

### Risks
 - None


miogalang:
:+1:

PR #219, opened 2015-11-04
[MI-331] Add option to set additional headers to the client

joseconsador:
Add option to set additional headers to the client. Initial use-case for this is to keep track of requests made per integration.

/cc @zendesk/mintegrations @mmolina @miketineo

### References
 - Jira link: https://zendesk.atlassian.net/browse/MI-331

### Risks
 - Medium: default headers could be overwritten and unexpected responses/requests could be sent.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -47,11 +47,11 @@ class Http
             $options
         );
 
-        $headers = [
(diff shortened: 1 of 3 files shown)


jwswj:
:+1:

miogalang:
:+1:

PR #227, opened 2015-12-09
Fix the route for finding a trigger

miogalang:
Fix the route to use when finding a trigger

/cc @zendesk/mintegrations @mmolina @miketineo

### References
 - Jira link: 

### Risks
 - None


miogalang:
This is wrong i was looking at zopim

PR #229, opened 2016-01-06
Fix Audit Logs filtering for correct filter parameter specification

ggrossman:
The documentation for Audit Logs (https://developer.zendesk.com/rest_api/docs/core/audit_logs) incorrectly used to state that filter parameters looked like this:

```
?filter["source_type"]=user
?filter["source_type"]=rule
?filter["created_at"]=2013-01-01%2000:00:00,2013-01-01%2023:59:59
```

Based on that information, the PHP code was enforcing this format with the double-quotes present.

However, the API doesn't actually work like that!

What actually works looks like this:

```
By source type:
?filter[source_type]=user
?filter[source_type]=rule

By exact source:
?filter[source_type]=user&filter[source_id]=123

By actor:
?filter[actor_id]=123

By IP address:
?filter[ip_address]=127.0.0.1

By time of creation:
?filter[created_at][]=2013-01-01T00:00:00Z&filter[created_at][]=2013-01-01T23:59:59Z

Control sort order (default is `sort_by=created_at`, `sort_order=desc`):
?sort_by=actor_id&sort_order=asc
```

### DIFF
--- a/tests/Zendesk/API/UnitTests/Core/AuditLogsTest.php
+++ b/tests/Zendesk/API/UnitTests/Core/AuditLogsTest.php
@@ -15,6 +15,8 @@ class AuditLogsTest extends BasicTest
     public function testFindAll()
     {
         $queryParams = [
+            'sort_by'              => 'actor_id',
+            'sort_order'           => 'desc',
             'filter[source_type]'  => 'rule',
             'filter[valid]'        => 'somerule',
             'filter[created_at][]' => '2016-01-01T00:00:00Z'
--- a/tests/Zendesk/API/UnitTests/Core/AuditLogsTest.php
+++ b/tests/Zendesk/API/UnitTests/Core/AuditLogsTest.php
@@ -26,7 +26,7 @@ class AuditLogsTest extends BasicTest
         // We also expect url encoded keys and values
         $expectedQueryParams = [];
         foreach ($queryParams as $key => $value) {
-            $expectedQueryParams = array_merge($expectedQueryParams, [urlencode($key) => /*urlencode*/($value)]);
+            $expectedQueryParams = array_merge($expectedQueryParams, [urlencode($key) => $value]);
         }
 
         $this->assertEndpointCalled(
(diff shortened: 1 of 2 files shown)


tests/Zendesk/API/UnitTests/Core/AuditLogsTest.php
```diff
@@ -15,15 +15,18 @@ class AuditLogsTest extends BasicTest
...
+            'filter[valid]'        => 'somerule',
+            'filter[created_at][]' => '2016-01-01T00:00:00Z'
         ];
 
         // We expect invalid parameters are removed.
         // We also expect url encoded keys and values
         $expectedQueryParams = [];
         foreach ($queryParams as $key => $value) {
-            $expectedQueryParams = array_merge($expectedQueryParams, [urlencode($key) => urlencode($value)]);
+            $expectedQueryParams = array_merge($expectedQueryParams, [urlencode($key) => /*urlencode*/($value)]);
```
ggrossman: The `Uri::withQueryValue` method called by Http.php seems to urlencode the key but not the value of query params. Any ideas?

miogalang: Hi @ggrossman, We traced it back to the implementation of guzzle. It seems they really aren't encoding ':' characters.

I tried it live though and it works. So we should be fine even if this is not encoded.

We did find a bit of a problem with passing the created at parameter. Zendesk wants you to pass 2 parameters with the same key, which guzzle does not allow. So we should re-think how to handle that. I will bring it up with the team.

ggrossman: The two `filter[created_at][]` parameter thing is weird and I'd be happy if we moved away from it in the Zendesk API, but it is what happens to work today.

miogalang:
:+1: @joseconsador Please check this as well

joseconsador:
:+1:

miogalang:
Thanks a lot @ggrossman!

pdeuter:
:chicken:

PR #232, opened 2016-02-23
Add CreateOrUpdate as a new way of updating Users

eski009:
Currently there is no way to use the createOrUpdate API call.

### DIFF
--- a/src/Zendesk/API/Resources/Core/Users.php
+++ b/src/Zendesk/API/Resources/Core/Users.php
@@ -39,6 +39,7 @@ class Users extends ResourceAbstract
         findMany as traitFindMany;
     }
     use UpdateMany;
+    use CreateOrUpdateMany;
 
     /**
      * @var UserIdentities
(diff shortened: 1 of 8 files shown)


joseconsador:
@eski009 Thanks for this, there are some errors on the unit test though. If you could fix them on your end, we'd be happy to re-review this PR.

src/Zendesk/API/Resources/Core/Users.php
```diff
@@ -39,6 +39,7 @@ class Users extends ResourceAbstract
         findMany as traitFindMany;
     }
     use UpdateMany;
+    use CreateOrUpdateMany;
```
miogalang: This seems to be throwing an error. I think a `use` is missing up top

eski009:
@miogalang  - thanks. Have updated

eski009:
@joseconsador - should be fixed.

src/Zendesk/API/Traits/Resource/CreateOrUpdateMany.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace Zendesk\API\Traits\Resource;
+
+use Zendesk\API\Exceptions\RouteException;
+use Zendesk\API\Http;
+
+/**
+ * Allows resources to call a bulk show endpoint.
```
joseconsador: Needs updating.

joseconsador:
:+1: after @eski009 adds the comment typo fix

miogalang:
Some tests needed for this but that can be done by us. Escalating the writing of unit tests for this to Jira.

miogalang:
:+1: thanks @eski009

PR #234, opened 2016-03-05
Fixed returning type declaration

TemirkhanN:
json_decode() returns an object if second argument is not being set to true.


miogalang:
:+1:

joseconsador:
:+1:

miogalang:
Thanks @TemirkhanN

PR #237, opened 2016-03-11
Fixed index of parameters for tags

antoscarface:
According to the parameter to pass to "tags" request:

```
curl https://{subdomain}.zendesk.com/api/v2/tickets/{id}/tags.json \
  -X PUT -d '{ "tags": ["customer"] }' \
  -H "Content-Type: application/json" -v -u {email_address}:{password}
```

what must be passed should be in this format: `{ "tags": ["customer"] }`, but the name of object is singularized by ResourceAbstract.
With this change, I force the object name to not be singularized and it works great now!


joseconsador:
@antoscarface Thanks for this. The travis test failed though due to coding standard restrictions, you can easily fix this by doing `vendor/bin/phpcbf --standard=PSR2 --extensions=php -w src` . Also, if you could add some live tests to verify this fix that would be much appreciated.

jmramos02:
hi @antoscarface any updates on this?

antoscarface:
PR Updated:
* Fixed phpcs errors
* Added tests

Sorry that I forgot this PR.
I hope is good now :)

PR #238, opened 2016-03-15
added helpdesk articles to the api

kaioken:
Added articles to the helpdesk api calls

Thanks


joseconsador:
@kaioken Thanks for this. You've committed changes to file permissions though. Could you update this PR but with `git config core.fileMode false` ? Thanks

kaioken:
@joseconsador will do, thanks

Issue #241, opened 2016-04-13
create_or_update (single) method missing


devanes:
Hello,

It would be nice if the create_or_update API method (single) would be available as well, for example for a user. You have create, update, and create_or_update_many. Of course the latter can be used, but having create_or_update as well would be nice, that way we don't have to check the job_status to see whether the action has succeeded.

Regards, Martijn

PR #244, opened 2016-04-27
Add create or update

jmramos02:
/cc @zendesk/mintegrations  @miogalang @joseconsador 

### Description

Add create and update trait

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-639

### Risks
* None


src/Zendesk/API/Resources/Core/Tickets.php
```diff
@@ -181,6 +181,31 @@ public function createFromTweet(array $params)
...
+     * Create or Update a ticket
+     *
+     * @param array $params
+     *
+     * @throws MissingParametersException
+     * @throws ResponseException
+     * @throws \Exception
+     * @return mixed
+     */
+    public function createOrUpdate(array $params)
```
miogalang: Is it not possible to re-use the trait here?

jmramos02:
@miogalang will remove that and use the trait createOrUpdate implementation

src/Zendesk/API/Traits/Resource/CreateOrUpdate.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Zendesk\API\Traits\Resource;
+
+use Zendesk\API\Exceptions\RouteException;
+
+trait CreateOrUpdate
```
miogalang: Is this only available for tickets? what else has this?

jmramos02:
Users has it, https://developer.zendesk.com/rest_api/docs/core/users#create-or-update-user

miogalang:
@jmramos02 Would be great to have it in all resources that have it. Then you can write a generic unit test for the trait once.

jmramos02:
Okay noted. will rewrite the ticket implementation

jmramos02:
Correction. tickets does not implement create_or_update https://github.com/zendesk/zendesk_api_client_php/issues/241 refers to "users" and not "tickets"

miogalang:
So the create_or_update method is only available for users and nothing else? is that correct?@jmramos02

jmramos02:
Yes @miogalang

miogalang:
@jmramos02 I don't see the point in pulling it out into a trait if it is only used once. We should just keep it in users if it is only available for one. Then think of pulling it out in a trait if we will re use it. This feels like unnecessary optimisation.

jmramos02:
noted @miogalang

src/Zendesk/API/Resources/Core/Users.php
```diff
@@ -9,6 +9,7 @@
 use Zendesk\API\Resources\ResourceAbstract;
 use Zendesk\API\Traits\Resource\CreateMany;
 use Zendesk\API\Traits\Resource\CreateOrUpdateMany;
+use Zendesk\API\Traits\Resource\CreateOrUpdate;
```
miogalang: I think we don't need this anymore

tests/Zendesk/API/UnitTests/Core/UsersTest.php
```diff
@@ -169,4 +169,17 @@ public function testChangePassword()
             $this->client->users($userId)->changePassword($postFields);
         }, "users/{$userId}/password.json", 'PUT');
     }
+
+    /*
+     * Tests if the createOrUpdate function calls the correct endpoint and passes the correct POST data
+     */
+    public function testCreateOrUpdate()
+    {
+        $postFields = ['id' => '1859295', 'name' => 'John Smith'];
```
miogalang: Please use Faker for these values

miogalang:
:+1: good work

joseconsador:
👍

pdeuter:
:chicken:

PR #246, opened 2016-05-24
Retrieve articles

jmramos02:
/cc @zendesk/mintegrations @miogalang @joseconsador

### Description
Add retrieval of articles

### Dependencies
* None

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-421

### Risks
* [medium] Helpdesk may not work


src/Zendesk/API/Resources/HelpCenter.php
```diff
@@ -36,7 +37,8 @@ public function __construct(HttpClient $client)
     public static function getValidSubResources()
     {
         return [
-            'categories' => Categories::class
+            'categories'    => Categories::class,
+            'articles'      => Articles::class
```
miogalang: add a , at the end so it looks nicer

src/Zendesk/API/Resources/HelpCenter/ResourceAbstract.php
```diff
@@ -18,4 +18,31 @@ protected function getResourceNameFromClass()
 
         return 'help_center/' . $resourceName;
     }
+
```
miogalang: Is this duplicated code from the other abstract?

src/Zendesk/API/Traits/Resource/Localize.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
miogalang: I think the name should be `Locales` ? what do you think?

jmramos02: I think it is, renaming..

src/Zendesk/API/Resources/HelpCenter/ResourceAbstract.php
jmramos02: Yes it is

miogalang: We should try not to replicate code, and keep it DRY. Maybe pull this out into a trait? do they share any other methods?

src/Zendesk/API/Resources/HelpCenter/ResourceAbstract.php
```diff
@@ -18,4 +18,31 @@ protected function getResourceNameFromClass()
...
+    /**
+     * Generate a route depending on a localization set
+     * @param string $name
+     * @param array $params
+     */
+    public function getRoute($name, array $params = [])
+    {
+        $routesWithLocale = ['findAll', 'find', 'create', 'update'];
```
joseconsador: @jmramos02 What would happen if there are resources that don't use all of these?

jmramos02: Will pull it out on a trait so it can be optional

jmramos02:
💇 @miogalang @joseconsador

src/Zendesk/API/Resources/HelpCenter/Articles.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @{inheritdoc}
+     */
+    protected $objectName = 'article';
+
+    /**
+     * @var locale
+     */
+    private $locale;
```
miogalang: Please move this to the locales trait

jmramos02:
💇 @miogalang

miogalang:
👍 Good job

joseconsador:
👍

PR #251, opened 2016-07-23
use example.com for sample codes

DQNEO:
company.com exists in reality.
https://whois.icann.org/en/lookup?name=company.com

we should use example.com instead.
http://www.rfc-editor.org/rfc/rfc2606.txt


joseconsador:
👍 
/cc @zendesk/mintegrations

miogalang:
👍

PR #252, opened 2016-07-23
eliminate username property

DQNEO:
HttpClient->username is not used, so no need to pass $username on construction


joseconsador:
👍 

/cc @zendesk/mintegrations

miogalang:
👍

Issue #218, opened 2015-10-30
Documentation of return values


netbrothers-tr:
In some places the return annotations simply consist of the word "mixed". In my opinion, should it matter, this isn't enough.

Example: https://github.com/zendesk/zendesk_api_client_php/blob/v2.0.7/src/Zendesk/API/Resources/Core/Tickets.php#L191

miogalang:
:+1:  Agree. This should be something more descriptive. Will add this to the list of technical debt. Not sure how soon we can get to it though.

netbrothers-tr:
@miogalang Thanks for your response. I really appreciate the client and I will patiently wait for updates in this field.

miogalang:
Closing this. Solved with PR from https://github.com/zendesk/zendesk_api_client_php/pull/254

netbrothers-tr:
Thanks guys! :thumbsup:

PR #254, opened 2016-07-26
[MI-742] Resolve PHPdoc inconsistencies in resource classes

samgavinio:
/cc @zendesk/mintegrations

### Description

Update return types and method annotations in several resources.

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-742

### Risks
* [low] PHPdocs may be incorrect.


samgavinio:
@miogalang related to https://github.com/zendesk/zendesk_api_client_php/issues/218.

We could do `stdClass|null`. All this is coming from the call to json_decode.

miogalang:
👍

jmramos02:
👍

joseconsador:
👍

miogalang:
@samgavinio yeah we should do `stdClass|null` that would be better as it is more specific. `mixed` doesn't really tell you much.

samgavinio:
@miogalang I'll open another PR to address #218. Code base is mostly returning `mixed`, I'll do it separately.

PR #258, opened 2016-08-01
Use stdClass for return types of several methods of core resources classes.

samgavinio:
/cc @zendesk/mintegrations

### Description

Use `stdClass | null` for return types of several GET, POST, PUT methods of core resources classes 

Use `null` for return types of DELETE methods.

### Risks
* [low] PHPdoc return types may be incorrect.


miogalang:
👍  ❤️

joseconsador:
👍

jmramos02:
👍

PR #260, opened 2016-08-08
 Closes #221. Ticket creation via voice api.

### LINKED ISSUE #221: Zendesk Voice API support
There seems to be no access to the Zendesk Voice API. I am looking for a way to create voice tickets. Can it be done with this client? If yes, how? And if not, what would be the best practice with PHP.

mindreader:
Used the PhoneNumber as a model.  Added a small unit test, live tested in my own code.


mindreader:
Sorry I should not have pull requested this from my master branch.

PR #261, opened 2016-08-08
 Closes #221. Ticket creation via voice api.

### LINKED ISSUE #221: Zendesk Voice API support

mindreader:
Wrote a small unit test, ran phpcs on tests and source, ran in own code, and it seems to work.  Based off of PhoneNumber.


src/Zendesk/API/Resources/Voice/Tickets.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    protected function setUpRoutes()
+    {
+        $this->setRoute('create', "{$this->getResourceName()}.json");
+    }
+
+    public function create(array $params, $routeKey = __FUNCTION__)
```
joseconsador: Docbloc here and this should be 👍

src/Zendesk/API/Resources/Voice/Tickets.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Zendesk\API\Traits\Resource\Create;
+
+/**
+ * Class Tickets
+ * https://developer.zendesk.com/rest_api/docs/voice-api/voice_integration
+ */
+class Tickets extends ResourceAbstract
+{
+    use Create;
```
samgavinio: This trait doesn't seem to have use since it's immediately overridden in this resource.

mindreader:
Changes as requested.  Sorry for the delay.

PR #266, opened 2016-09-19
[MI-811] Create embeddable config_sets resource

samgavinio:
/cc @zendesk/mintegrations

### Description

Create API client resource for the embeddable config_set create (`POST: /embeddable/api/config_sets.json`) and update (`PUT: /embeddable/api/config_sets/{$id}.json`) endpoints.

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-811

### Risks
* [HIGH] This essentially touches all the API calls due to the modified api base path implementation

### DIFF
--- a/src/Zendesk/API/Resources/Embeddable/ResourceAbstract.php
+++ b/src/Zendesk/API/Resources/Embeddable/ResourceAbstract.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Zendesk\API\Resources\Embeddable;
+
+use Zendesk\API\Traits\Resource\ResourceName;
+
+/**
+ * Abstract class for Embeddable resources
+ */
+abstract class ResourceAbstract extends \Zendesk\API\Resources\ResourceAbstract
+{
+    use ResourceName;
+
+    /**
+     * @var string
+     **/
+    protected $prefix = '../../embeddable/api/';
+}
--- a/src/Zendesk/API/Resources/Embeddable/ResourceAbstract.php
+++ b/src/Zendesk/API/Resources/Embeddable/ResourceAbstract.php
@@ -14,5 +14,10 @@ abstract class ResourceAbstract extends \Zendesk\API\Resources\ResourceAbstract
     /**
      * @var string
      **/
-    protected $prefix = '../../embeddable/api/';
+    protected $prefix = 'embeddable/api/';
+
+    /**
+     * @var string
+     */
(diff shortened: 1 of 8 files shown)


jmramos02:
👍

src/Zendesk/API/Resources/Embeddable/ResourceAbstract.php
```diff
@@ -0,0 +1,18 @@
...
+ * Abstract class for Embeddable resources
+ */
+abstract class ResourceAbstract extends \Zendesk\API\Resources\ResourceAbstract
+{
+    use ResourceName;
+
+    /**
+     * @var string
+     **/
+    protected $prefix = '../../embeddable/api/';
```
miogalang: What is the reason for the `../..`? what does the endpoint look like?

tests/Zendesk/API/UnitTests/Embeddable/ConfigSetsTest.php
```diff
@@ -0,0 +1,41 @@
...
+use Zendesk\API\UnitTests\BasicTest;
+
+class ConfigSetsTest extends BasicTest
+{
+    /**
+     * Tests if the client can call and build the create config sets endpoint
+     */
+    public function testCreate()
+    {
+        $params = [
```
miogalang: can we use faker for this.

```
hexcolor 
randomElement($array = array ('a','b','c')) // 'b'
```

perhaps would be good to use

src/Zendesk/API/Resources/Embeddable/ResourceAbstract.php
samgavinio: The endpoint is `http://support.zd-dev.com/embeddable/api/config_sets.json`. It's different from the rest of the existing endpoints in that it isn't under `api/{version}/`.

samgavinio:
@miogalang 💇 

To allow overriding the `api/v2` api prefix, the recent change adds a protected variable for the `$apiBasePath` which defaults to `api/v2` in the `ResourceAbstract`. Derived resource classes may override this to set a different api base path. 

This variable is then passed to the `HttpClient` class using a setter method as seen [here](https://github.com/zendesk/zendesk_api_client_php/blob/1df6723d635113bc45c8a49e3ed76c59540bc32f/src/Zendesk/API/Resources/ResourceAbstract.php#L62)

PR #268, opened 2016-09-26
ApiResponseException::getErrorDetails now returns an array, like spec…

NinoSkopac:
…ified in the PHPDoc annotation

I guess this is a pretty big BC break, so it's smarter to just change annotation for this major/minor version.

### DIFF
--- a/src/Zendesk/API/Exceptions/ApiResponseException.php
+++ b/src/Zendesk/API/Exceptions/ApiResponseException.php
@@ -58,6 +58,6 @@ class ApiResponseException extends \Exception
      */
     public function getErrorDetails()
     {
-        return $this->errorDetails;
+        return json_decode($this->errorDetails, true);
     }
 }


src/Zendesk/API/Exceptions/ApiResponseException.php
```diff
@@ -58,6 +58,6 @@ public function __construct(RequestException $e)
      */
     public function getErrorDetails()
```
kcasas: It wouldn't be advisable to introduce a BC breaking change without updating the version. I think it would be better to just a way for the function to return an array. Change could be

` *
     * @return json|array
     */
    public function getErrorDetails($decode = false)
    {
        return $decode ? json_decode($this->errorDetails, true): $this->errorDetails;
    }`

Braunson:
Is this something that's going to be merged? The specs say it responds with an array but instead a string is returned in 2.x now.

ecoologic:
There's no appetite for introducing breaking changes at this point. We'll likely update the docs instead, as mentioned.

Thank you for your contribution, it would have been better to implement the code like you did in the first place, but now we don't want to force all consumers to update.

Issue #272, opened 2016-10-04
TicketComments is not usable


NinoSkopac:
Not imported in `src/Zendesk/API/HttpClient.phpsrc/Zendesk/API/HttpClient.php`.

NinoSkopac:
I haven't tested it, but I'd assume the same problem is present for some other calls.

ghermans:
@NinoSkopac, @joseconsador  any update regarding this?

NinoSkopac:
@ghermans Ship it

ghermans:
@NinoSkopac ok, are you planning to make an example file in the samples folder?

NinoSkopac:
@ghermans Sure thing, will do it later today or tomorrow.

ghermans:
@NinoSkopac i added the example #287, so you can close this can be marked as solved i guess ?

PR #274, opened 2016-10-04
Made TicketComments usable

NinoSkopac:
Fixes https://github.com/zendesk/zendesk_api_client_php/issues/272


joseconsador:
@NinoSkopac the preferred way to use TicketComments would be `$client->tickets($ticketId)->comments()->findAll(). /cc @miogalang

Issue #269, opened 2016-09-29
Needs Documentation


fireproofsocks:
**There needs to be a minimum set of examples demonstrating how to use this package.**  It isn't at all obvious how the raw API endpoints detailed in the API docs at https://developer.zendesk.com/rest_api/docs/core/introduction are accessed or modified using the PHP classes and methods in the package.

"Simple" tasks like looking up a user by email etc. is a mystery that usually ends up in the support forum or winding through a maze of traits, inheritance, and magic methods in the source code that is difficult to navigate, even for seasoned developers.  Sometimes those expeditions yield a solution, sometimes they don't, but _they are always a needless waste of time_.

These are the things that should be included in the documentation, and unless I am missing something, they are not.  The README is bare, the Wiki is empty, and the irony is rich.  Attention in this area would be much appreciated.

NinoSkopac:
I've been using this SDK for the past week as I am implementing it into a support portal, and it's been a nightmare. No docs whatsoever. I've just spent the last 30 minutes figuring out which method name should I use for getting ticket comments. No luck so far.

NinoSkopac:
Furthermore, it's not just _missing_ docs, it's also _incorrect_ PHPDocs. See https://github.com/zendesk/zendesk_api_client_php/issues/267

ghermans:
@fireproofsocks i have just created a PR with some new samples and if i have the time i will add more soon.

joseconsador:
Changes from @ghermans merged. Closing this for now.

PR #282, opened 2016-10-30
#269 - Added new samples

ghermans:


ghermans:
Some files where missing, will create a new PR.

PR #293, opened 2016-11-29
Enable anonymous access

kcasas:
/cc @zendesk/mintegrations

### Description

Enable anonymous access. Added missing return on some functions.

### Risks
[medium] May break functionalities that expects a thrown exception when auth values are not supplied


tests/Zendesk/API/UnitTests/Core/AuthTest.php
```diff
@@ -12,6 +16,23 @@
 class AuthTest extends BasicTest
 {
     /**
+     * Test if request is still sent even without auth details
+     */
+    public function testAnonymousAccess()
+    {
+        $client = $this
```
samgavinio: This test doesn't seem like it's testing the anonymous access behaviour directly or at least it isn't too obvious. Shouldn't there at least be an assertion that `$client->getAuth()->prepareRequest()` is not called?

Also there's a `client` class property which contains an instance of `Zendesk\API\HttpClient` already, may be you can use that.

kcasas: I believe there is no assertion for method calls. If there is, kindly refer it to me. I've tried using that client but it seems that it is not a mocked object.

miogalang: you can do 

`$client->expects('getAuth')->never()`

src/Zendesk/API/Http.php
```diff
@@ -82,7 +82,10 @@ public static function send(
         }
 
         try {
-            list ($request, $requestOptions) = $client->getAuth()->prepareRequest($request, $requestOptions);
+            // enable anonymous access
+            if ($client->getAuth()) {
+                list ($request, $requestOptions) = $client->getAuth()->prepareRequest($request, $requestOptions);
```
joseconsador: `getAuth()` doesn't really do much and is not reused in the library. Perhaps we can move the implementation of this conditional to `Zendesk\API\Utilities\Auth::prepareRequest()` .

tests/Zendesk/API/UnitTests/Core/AuthTest.php
kcasas: getAuth is called once when checking if it exists or not.

samgavinio: 👍 on the client object not being a mocked object. I guess you can create your own mock but you should be able to assert the method call if you use mockery demeter chains:

http://docs.mockery.io/en/latest/reference/demeter_chains.html

miogalang: Maybe prepareRequest is what you should check as not being called then.

src/Zendesk/API/Http.php
joseconsador: That's possible, the Auth class would no longer be needed if that's the case. Check out the [ruby implementation](https://github.com/zendesk/zendesk_api_client_rb/blob/master/lib/zendesk_api/client.rb#L158) for some inspiration.

Issue #289, opened 2016-11-08
 Error completing request


julzpalomiano:
I downloaded (https://github.com/zendesk/zendesk_api_client_php) using composer and copy paste the basic operation there but it seems that even if I inputted the correct subdomain, username, and token I always come up with the error saying that

Error completing request [url] https://xxx.zendesk.com/api/v2/tickets.json [http method] GET [body] in

joseconsador:
@julzpalomiano Could you paste the error here? It's cut-off.

julzpalomiano:
@joseconsador here you go..

Uncaught Zendesk\API\Exceptions\ApiResponseException: Error completing request [url] https://julz.zendesk.com/api/v2/tickets.json [http method] GET [body] in C:\xampp\htdocs\Dropbox\practice\zendeskapi\vendor\zendesk\zendesk_api_client_php\src\Zendesk\API\Http.php:89 Stack trace: #0 C:\xampp\htdocs\Dropbox\practice\zendeskapi\vendor\zendesk\zendesk_api_client_php\src\Zendesk\API\HttpClient.php(445): Zendesk\API\Http::send(Object(Zendesk\API\HttpClient), 'tickets.json', Array) #1 C:\xampp\htdocs\Dropbox\practice\zendeskapi\vendor\zendesk\zendesk_api_client_php\src\Zendesk\API\Traits\Resource\FindAll.php(31): Zendesk\API\HttpClient->get('tickets.json', Array) #2 C:\xampp\htdocs\Dropbox\practice\zendeskapi\mclink.php(32): Zendesk\API\Resources\Core\Tickets->findAll() #3 {main} thrown in 



salamat :)

joseconsador:
Can you give the output for:

```php
try {
  $response = $this->client->tickets()->findAll();
} catch (Zendesk\API\Exceptions\ApiResponseException $e) {
  var_dump($e->getMessage());
}

```

julzpalomiano:
@joseconsador 

 "Error completing request [url] https://julz.zendesk.com/api/v2/tickets.json [http method] GET [body] "

phongtranh:
I have same error, can you tell me what its happend?

Polou:
Same problem here with php 5.6 : 
`Error completing request [url] https://eureciahelp.zendesk.com/api/v2/tickets.json?include=users,groups [http method] GET [body] `

puyaalexxx:
Hi! When are you going to fix this issue? thanks

[...]

Issue #279, opened 2016-10-27
zendesk/zendesk_api_client_php contains .git and .gitignore


placeposition:
I just installed with composer and this directory contains another .git directory which makes a mess of my parent application .git

placeposition:
/vendor/zendesk/zendesk_api_client_php

victoriafrench:
Composer is bringing down a file structure that contains a .git folder. This causes git to refuse to check in the files. The .git should never be distributed under any circumstances, they are per developer machine. Even after removing the .git there are still issues with committing and we are investigating.

karl-casas777:
@placeposition @victoriafrench  I bet the you installed the package in `dev-master` version. Any composer package installed in a `dev-master` version will contain the .git folder because it assumes you want to contribute to the package. Installing the release tags instead will not include the .git folder.

miogalang:
Please use the version tags for a more stable experience. Closing this issue.

victoriafrench:
Perhaps you should consider updating your readme to explicitly state how to consume the project instead of how to contribute to it. Neither scenario is appropriately documented. On my end it won't matter much, I convinced my client to abandon this integration. I will also reiterate that there is NEVER a reason to ship a `.git` directory... ever. period.

kcasas:
@victoriafrench You are right that's one of the reason why dev-master is not encouraged. Do know we don't intend for this behaviour. This happens with every composer dependency that has dev-master on it, not just this. I do hope you keep supporting this integration.

miogalang:
Hi @victoriafrench 

I agree with you. This is default behaviour of composer for the master branch though and really there is nothing we can do about it.

What I would suggest is to install the library using the command

`composer require zendesk/zendesk_api_client_php` as this will update your composer.json automatically with the latest version.

This will also add the integration in your vendor folder without the git.

Thank you for the feedback we will be updating the readme in the next few days with this change in install instructions. 

Please let us know if there is anything else we can help you with.

Issue #264, opened 2016-08-30
Error completing request [URL] --- [http method] GET [body]


ZeusBeard:
When trying to access the Zendesk API it keeps failing, i can use guzzle to access the rest API fine, it just appears to be whenever i try and use the directed method.

You can see my example below

![image](https://cloud.githubusercontent.com/assets/21336008/18088452/ae0f4102-6eb2-11e6-9058-91d6503e428f.png)

druizfr:
Hi Zeus,
I'm having the same issue over here. Did you manage to solve it?

druizfr:
Hi Zeus,
Removing the trys it return me the real issue. Error 60 in cURL and fixed it with this: http://stackoverflow.com/questions/29822686/curl-error-60-ssl-certificate-unable-to-get-local-issuer-certificate

Issue #248, opened 2016-06-13
Error completing request [url] https://my22.zendesk.com/api/v2/tickets.json [http method] GET [body]


undentifiedkc:
Hi, I was using the this API on **localhost** to try retrieving data, but error encounter after i fill in the credential.

Please find below for the error.

Uncaught exception 'Zendesk\API\Exceptions\ApiResponseException' 
with message 'Error completing request [url] https://my22.zendesk.com/api/v2/tickets.json [http method] GET [body] ' 
in C:\xampp2\htdocs\testing\zendesk_api_client_php-master\src\Zendesk\API\Http.php:89 
Stack trace: 
    #0 C:\xampp2\htdocs\testing\zendesk_api_client_php-master\src\Zendesk\API\HttpClient.php(400): Zendesk\API\Http::send(Object(Zendesk\API\HttpClient), 'tickets.json', Array) 
    #1 C:\xampp2\htdocs\testing\zendesk_api_client_php-master\src\Zendesk\API\Traits\Resource\FindAll.php(34): Zendesk\API\HttpClient->get('tickets.json', Array) 
    #2 C:\xampp2\htdocs\testing\zendesk_api_client_php-master\samples\getTickets.php(16): Zendesk\API\Resources\Core\Tickets->findAll() #3 {main} thrown in C:\xampp2\htdocs\testing\zendesk_api_client_php-master\src\Zendesk\API\Http.php on line 89

I have already enable token access in the domain > settings > API and copy the token from there.

joseconsador:
@undentifiedkc Could you paste the code that you use here?

julzpalomiano:
same issue as mine..

joseconsador:
@julzpalomiano @undentifiedkc Can you catch the exception and dump the response here? Thanks.

samgavinio:
For any one still experiencing this issue, this is often due to a curl misconfiguration in your host machine. We've made a change to this client's exception handling which should give you more context about the ApiResponseException you are getting.

Please update to at least v2.1.10 and inspect the context of the exception, as it should help you better debug the issue with your curl set-up.

As this is once of those older issues, we'll be closing this.

PR #296, opened 2016-12-09
Outputs message from exception and updated README

kcasas:
/cc @zendesk/mintegrations

### Description

There has been a lot of complaints of error requests in examples. Was not able to replicate the problem as it may be a development setup issue. This outputs the actual message from the thrown exception to help users debug the issue. Also updated readme on how to install using composer.

### References
* Composer installation problem https://github.com/zendesk/zendesk_api_client_php/issues/279
* Error request completion https://github.com/zendesk/zendesk_api_client_php/issues/289, https://github.com/zendesk/zendesk_api_client_php/issues/264, and https://github.com/zendesk/zendesk_api_client_php/issues/248

### Risks
* none

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,13 @@ The Zendesk PHP API client can be installed using [Composer](https://packagist.o
 
 ### Composer
 
-Inside of `composer.json` specify the following:
+To install run `composer require zendesk/zendesk_api_client_php` or
+inside of `composer.json` specify the following:
 
 ``` json
 {
   "require": {
-    "zendesk/zendesk_api_client_php": "dev-master"
+    "zendesk/zendesk_api_client_php": "^2.1"
   }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -23,16 +23,7 @@ The Zendesk PHP API client can be installed using [Composer](https://packagist.o
 
 ### Composer
 
-To install run `composer require zendesk/zendesk_api_client_php` or
-inside of `composer.json` specify the following:
-
-``` json
-{
-  "require": {
-    "zendesk/zendesk_api_client_php": "^2.1"
-  }
-}
-```
+To install run `composer require zendesk/zendesk_api_client_php`
 
 ### Upgrading from V1 to V2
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -23,12 +23,13 @@ The Zendesk PHP API client can be installed using [Composer](https://packagist.o
...
 
-Inside of `composer.json` specify the following:
+To install run `composer require zendesk/zendesk_api_client_php` or
+inside of `composer.json` specify the following:
 
 ``` json
 {
   "require": {
-    "zendesk/zendesk_api_client_php": "dev-master"
+    "zendesk/zendesk_api_client_php": "^2.1"
```
miogalang: maybe we shouldn't just get rid of this part completely. this is something we would need to update every version release. 

I think composer require zendesk/zendesk_api_client_php is enough

see: https://github.com/fzaninotto/Faker https://github.com/guzzle/guzzle

these are very popular libraries which only use require as install instructions

samples/tickets/getTicketComments.php
```diff
@@ -24,5 +24,6 @@
     print_r($tickets);
     echo "</pre>";
 } catch (\Zendesk\API\Exceptions\ApiResponseException $e) {
+    echo $e->getMessage().'</br>';
     echo 'Please check your credentials. Make sure to change the $subdomain, $username, and $token variables in this file.';
```
miogalang: Maybe we should get rid of this message since we are already showing the error from the API?

samples/tickets/createTicketWithAttachment.php
```diff
@@ -45,5 +45,6 @@
     print_r($newTicket);
     echo "</pre>";
 } catch (\Zendesk\API\Exceptions\ApiResponseException $e) {
+    echo $e->getMessage().'</br>';
```
samgavinio: Shouldn't we be using `$e->getErrorDetails()` instead? It has better information of what actually is going on with the Guzzle exception as @miogalang has found here: https://github.com/zendesk/zendesk_api_client_php/issues/220#issuecomment-155713049

kcasas: @samgavinio Tried it. It turns out that `getErrorDetails` is just a subset of `getMessage`.

samples/tickets/getTicketComments.php
kcasas: @miogalang I think the message is still helpful as those variables are setup in each sample file and can easily be ignored.

kcasas:
@miogalang 💇

Issue #299, opened 2016-12-21
Async ticket creation support


FuzzyTree:
According to the documentation, it's possible to create a ticket asynchronously. But I couldn't find this feature in the sdk. Is it supported?

https://developer.zendesk.com/rest_api/docs/core/tickets#create-ticket

> Ticket creation can sometimes take a while because of complex business rules. Asynchronous creation allows you to get the response back quickly and queues a background job to do the actual work.
> 
> To make an asynchronous request, simply add the parameter async in your request.
> 
> POST api/v2/tickets.json?async=true
> 
> As soon as the request is received, it returns a 202 Accepted. The response includes the new ticket id with a job_status JSON object. However, you may not be able to fetch the ticket until the job is done.

miogalang:
solved: https://github.com/zendesk/zendesk_api_client_php/pull/302

Available as of v2.1.7

FuzzyTree:
@miogalang Thank you!

PR #300, opened 2016-12-21
Add support for creating tickets asynchronously

FuzzyTree:
This PR makes it possible to create a ticket asynchronously by setting the `async` key to true in the array passed to tickets->create i.e.

        $ticket = [
            'async' => true,
            'subject' => 'Unit Test ' . rand(0,999999),
            'priority' => 'normal',
            'comment' => [
                'body' => 'Live Chat started.',
            ],
            'requester' => [
                'name' => 'PHPUnit',
                'email' => 'phpunit@test.com',
            ],
            'tags' => $all_tags,
            'via' => [
                'channel' => 'chat',
            ],
        ];

        $zendesk_response = Zendesk::tickets()->create($ticket);

Addresses #299

### DIFF
--- a/src/Zendesk/API/Traits/Resource/Create.php
+++ b/src/Zendesk/API/Traits/Resource/Create.php
@@ -24,6 +24,12 @@ trait Create
             }
 
             $route = $this->resourceName . '.json';
+            
+            if (isset($params['async']) && $params['async'] === true) {
+                $route .= '?async=true';
+                unset($params['async']);
+            }
+            
             $this->setRoute(__FUNCTION__, $route);
         }
 


miogalang:
@FuzzyTree The way to do this is to add to the async parameter to the query parameters.

```
    $ticket = [
        'queryParams' => [
            'async' => true,
        ],
        'subject' => 'Unit Test ' . rand(0,999999),
        'priority' => 'normal',
        'comment' => [
            'body' => 'Live Chat started.',
        ],
        'requester' => [
            'name' => 'PHPUnit',
            'email' => 'phpunit@test.com',
        ],
        'tags' => $all_tags,
        'via' => [
            'channel' => 'chat',
        ],
    ];

    $zendesk_response = Zendesk::tickets()->create($ticket);
```

FuzzyTree:
@miogalang adding `'queryParams' => ['async' => 'true']` (also with `true` as a boolean) did not work for me. The returned object has all keys instead of only the `id` key

miogalang:
@FuzzyTree You are right. Seems there is no support for query parameters for post requests. 

There should probably some matching test for this.
@joseconsador Mind chiming in on what could be a better way of doing this?

FuzzyTree:
How about

    if (!empty($params['queryParams']) && is_array($params['queryParams'])) {
        $route .= '?' . http_build_query($params['queryParams']);
        unset($params['queryParams']);
    }

miogalang:
@FuzzyTree Tried my hand at it. I came up with https://github.com/zendesk/zendesk_api_client_php/pull/302

The problem with putting this in the trait is that the `async` option becomes available to all resources using the trait. I want to limit it to the Ticket resource as that is the only one that should be able to do async

PR #302, opened 2016-12-23
Allow async query parameter for creating tickets

miogalang:
/cc @zendesk/mintegrations

### Description

Allows the creation of tickets asynchronously.
ref: https://developer.zendesk.com/rest_api/docs/core/tickets#create-ticket-asynchronously

### References
* JIRA: 

### Risks
* medium. Ticket creation mught fail

### DIFF
--- a/src/Zendesk/API/HttpClient.php
+++ b/src/Zendesk/API/HttpClient.php
@@ -429,7 +429,7 @@ class HttpClient
      *
      * @return \stdClass | null
      * @throws \Zendesk\API\Exceptions\AuthException
-     * @throws Exceptions\ApiResponseException
+     * @throws \Zendesk\API\Exceptions\ApiResponseException
      */
     public function get($endpoint, $queryParams = [])
     {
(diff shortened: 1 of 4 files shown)


tests/Zendesk/API/UnitTests/Core/TicketsTest.php
```diff
@@ -178,6 +178,26 @@ public function testCreateWithAttachment()
             'postFields' => $postFields,
         ]);
     }
+    
+    /**
+     * Tests if the client can call and build the create ticket witch attachment endpoint and initiate the file upload
```
samgavinio: Description doesn't seem to match the test.

src/Zendesk/API/HttpClient.php
```diff
@@ -454,18 +455,22 @@ public function get($endpoint, $queryParams = [])
      * @param       $endpoint
      * @param array $postData
      *
-     * @return \stdClass | null
+     * @param array $options
+     * @return null|\stdClass
+     * @throws \Zendesk\API\Exceptions\AuthException
      * @throws Exceptions\ApiResponseException
```
samgavinio: Should this also be changed to `\Zendesk\API\Exceptions\ApiResponseException`?

PR #303, opened 2017-01-05
Added functionality to showMacro when given a macro ID

marcoleary:
Added a method to return the macro object from an ID. I didn't know what the method should be called, though... When doing the same action with tickets the method is `find()` but the documentation suggested it should be `showMacro()`. I went with the latter but will happily change if recommended.

### DIFF
--- a/tests/Zendesk/API/UnitTests/Core/MacrosTest.php
+++ b/tests/Zendesk/API/UnitTests/Core/MacrosTest.php
@@ -3,6 +3,8 @@
 namespace Zendesk\API\UnitTests\Core;
 
 use Zendesk\API\UnitTests\BasicTest;
+use Zendesk\API\Exceptions\MissingParametersException;
+
 
 /**
  * Macros test class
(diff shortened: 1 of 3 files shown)


tests/Zendesk/API/UnitTests/Core/MacrosTest.php
```diff
@@ -14,6 +14,19 @@ class MacrosTest extends BasicTest
      * Test the `GET /api/v2/macros/active.json` endpoint
      * Lists active macros for the current user
      */
+    public function testShowMacro()
+    {
+        $id = 1;
```
jmramos02: Can you use faker for this? :)

jmramos02: Can this be randomly generated? :)

jmramos02:
@marcoleary looks great! can we have a another unit test wherein it checks for the exception thrown? :)

tests/Zendesk/API/UnitTests/Core/MacrosTest.php
jmramos02: Can this be randomly generated? :)

miogalang:
@marcoleary I'd actually prefer to use `find` for the method name for consistency with other resources.

tests/Zendesk/API/UnitTests/Core/MacrosTest.php
miogalang: What you can do here is 

above you add:
```
use Faker\Factory;
```

then that will allow you to do 
```
        $faker = Factory::create();
        $id = $faker->randomNumber();
```

marcoleary:
All done! Thanks for the feedback.

miogalang:
@marcoleary Actually, I looked into this closer. There actually already is a find for Macros existing in the Defaults Trait.

You should currently be able to do 

```
$client->macros->find($id)
```

PR #309, opened 2017-02-21
Fix user's merge function

### LINKED ISSUE #308: Unable to merge users
In [Zendesk\API\Resources\Core\Users](https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Resources/Core/Users.php) exists only one route for merge - users/me/merge.json.
There are no route with [this method](https://developer.zendesk.com/rest_api/docs/core/users#merge-users) for merging users

shershennm:
Tried to fix #308 without breaking backward compatibility

### DIFF
--- a/tests/Zendesk/API/UnitTests/Core/UsersTest.php
+++ b/tests/Zendesk/API/UnitTests/Core/UsersTest.php
@@ -46,7 +46,7 @@ class UsersTest extends BasicTest
         ];
 
         $this->assertEndpointCalled(function () use ($postFields) {
-            $this->client->users('me')->merge($postFields);
+            $this->client->users()->merge($postFields);
         }, 'users/me/merge.json', 'PUT', ['postFields' => ['user' => $postFields]]);
     }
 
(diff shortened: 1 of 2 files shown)


joseconsador:
@shershennm Looks good, can we have a test where {id} is replaced in the call? Thanks for this.

shershennm:
@joseconsador done

tests/Zendesk/API/UnitTests/Core/UsersTest.php
```diff
@@ -51,6 +51,50 @@ public function testMerge()
...
+            'id' => $userId,
+            'mergingId' => $mergingId,
+        ];
+
+        $postFields = [
+            'id' => $mergingId,
+        ];
+
+        $this->assertEndpointCalled(function () use ($postFields, $params, $userId) {
+            $this->client->users()->merge($params);
```
samgavinio: @joseconsador is normal for the client to support both?

* $this->client->users($id1)->merge(['mergingId'=> $id2]);
* $this->client->users()->merge(['id' => $id1, 'mergingId' => $id2]);

The ticket merge seems to only make use of the first approach. In any case looks good to me, we should probably add this to the samples as well. 👍

shershennm:
@joseconsador , @samgavinio Almost all functions accepts double behavior: passing `id` as chained parameter, or passing `id` to function parameters. Ticket merge also allows both ways :
- `$this->client->tickets()->(['id' => 12345, 'ids' => [ 123456 ] ]);`
- `$this->client->tickets(12345)->(['ids' => [ 123456 ] ]);`
But if we will use use `id` as `mergingId`, it breaking this "style", and causing mismatch in parameters. Maybe will be better to separate admin and user merges?

bretticus:
Currently, with v2.1.8, the API request requires:

`['user' => ['id' => {some id}]]`

However, when you send that as the merge parameter, it currently only gets by the hasKeys check...

```
if (! $this->hasKeys($params, $hasKeys)) {
            throw new MissingParametersException(__METHOD__, $hasKeys);
        }
```

If it's entered as `['id' => {some id}]`

However, that results in a Zendesk\API\Exceptions\ApiResponseException

bretticus:
I tried commenting the hasKeys check out and I just realized that it still gets sent using the `/me/` form anyway (hence this pull request!)

Thanks for fixing this!

For now, I'll just call the URL via cURL.

bretticus:
So far, I have tried:

```
$response = Http::send($client, sprintf('users/%d/merge.json', $id1),
							['postFields' => ['user' => ['id' => $id2]], 'method' => 'PUT']
						);
```
But now, I'm getting `422 Unprocessable Entity`

I can make the request via curl in console command. So strange!

joseconsador:
@shershennm 

To avoid confusion, @samgavinio 's suggestion would be more inline with the API docs and this client's usage. `$this->client->users($id1)->merge(['id' => $id2]);` . 

Given that, the `id` parameter on the merge method should be referring to the ID of the user to merge.
If no ID is provided when invoking `$this->client->users()` , `/me` should be used, which I think already is the case.

***So to summarize, let's only allow:***
* `$this->client->users($id1)->merge(['id' => $id2]);`
   - `curl -v -u {email_address}:{password} https://{subdomain}.zendesk.com/api/v2/users/{$id1}/merge.json \
  -H "Content-Type: application/json" -X PUT -d '{"user": {"id": $id2}}'`

* `$this->client->users()->merge(['id' => $id1]);`
  - `curl -v -u {email_address}:{password} https://{subdomain}.zendesk.com/api/v2/users/me/merge.json \
  -H "Content-Type: application/json" -X PUT -d '{"user": {"id": $id1}}'`

/cc @samgavinio thoughts?

PR #310, opened 2017-02-23
Remove getContents()

christopheg:
In the documentation from Guzzle, it is indicated that you can simply cast the getBody() to a string to use the response. The method getContents() will only read the remaining contents if a previous read() is performed.
see: http://docs.guzzlephp.org/en/latest/quickstart.html?#using-responses

This causes problems if you install a Guzzle Log middleware. This will read the body causing getContents() te become empty.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -93,7 +93,7 @@ class Http
         } finally {
             $client->setDebug(
                 $request->getHeaders(),
-                $request->getBody()->getContents(),
+                $request->getBody(),
                 isset($response) ? $response->getStatusCode() : null,
                 isset($response) ? $response->getHeaders() : null,
                 isset($e) ? $e : null
@@ -108,6 +108,6 @@ class Http
 
         $client->setSideload(null);
 
-        return json_decode($response->getBody()->getContents());
+        return json_decode($response->getBody());
     }
 }


samgavinio:
@christopheg Thanks. Doesn't look like anything will break but can you give us more context of how you plan on making use of the response body within your middleware?

From the looks of how it was previously, you could always just call `$response->getBody()` from within your middleware and it will still return the entire stream.

christopheg:
In order to be able to debug the communication, we want to have all HTTP traffic logged in a file. The Zendesk client doesn't offer this functionality so I was looking in the code. It seems the Zendesk client uses Guzzle for HTTP communication.
Guzzle has a very good way of logging the communication and the Zendesk client offers a way of passing a Guzzle client via the constructor. So the only thing I need to do is to pass a customized Guzzle client via the Zendesk client constructor. Like this:

```
$stack = \GuzzleHttp\HandlerStack::create();
$stack->push(
	\GuzzleHttp\Middleware::log(
		new Log(),  // Custom Log class
		new \GuzzleHttp\MessageFormatter(\GuzzleHttp\MessageFormatter::DEBUG)
	), 'log'
);

$guzzle_client = new \GuzzleHttp\Client(
	[
		'handler' => $stack,
	]
);

$api = new \Zendesk\API\HttpClient($ZENDESK_SUBDOMAIN, $ZENDESK_USERNAME, 'https', 'zendesk.com', 443, $guzzle_client);
```

The problem is that the Guzzle MessageFormatter already performs a read operation on the body. The result is that the Zendesk API client can no longer use $response->getBody()->getcontents() as getcontents() only returns the remaing content of the body (which is nothing)

samgavinio:
@christopheg Thanks for the context. Released as v2.1.8.

PR #313, opened 2017-03-08
bugfix: use the current version information

dexen:
HttpClient::VERSION was not updated during last few releases.
Just fetch the version number from VERSION file.


samgavinio:
@dexen Our bad, the class constant hasn't been updated during recent version bumps. 

We decided to stick with the class constant since reading the filesystem every time for the version seems like unnecessary overhead. We've been updating the stale version now and will make sure to keep it updated.

Thanks.

Issue #316, opened 2017-03-21
Error completing request [url] https://xxxxx.zendesk.com/api/v2/tickets.json [http method] POST [body] 


ghost:
Hello,

today all of a sudden all API calls failed with the this type of error on different endpoints:
Error completing request [url] https://xxxxx.zendesk.com/api/v2/tickets.json [http method] POST [body] 

Not sure if it had anything to do with the roll out earlier today (nhanced Security for Zendesk Single Sign-on), but updating the cert helped. Unfortunately we did experience downtime because of this.

Jeroen

kcasas:
Hi @jvandersandt can you elaborate more on what cert you updated? Is that the certificate used by the server for ssl connections?

jvandersandt:
Hello,

unfortunately the issue returned, I took the solution from  #297, but leaving it out and simply restarting apache removes the error again (for now at least).

Thanks,

kcasas:
So we can conclude that the problem isn't with the API endpoints but something to do with the environment you're using?

vetheslav:
Hello! I have the same problem.
You can watch my sentry issue: https://sentry.io/share/issue/3134373436332e323337303833303431/
and some values of code:
![2017-03-23 13 21 26](https://cloud.githubusercontent.com/assets/17473642/24243322/c651fee8-0fcb-11e7-8341-d57d21f204cf.png)

kcasas:
@vetheslav It seems like you're request body is empty. It's quite weird cause its supposed to throw a 422 status code plus descriptive errors about the required fields. Can you give us the snippet for `opt/asterisk_aon/src/ZendeskLib.php`? Thanks.

vetheslav:
@kcasas yes.
Here is my data:
![2017-03-24 11 32 16](https://cloud.githubusercontent.com/assets/17473642/24286387/0b368a6e-1086-11e7-8ec2-7a5ec24e55a7.png)
Code of opt/asterisk_aon/src/ZendeskLib.php: http://codepad.org/6RyK4Ama

samgavinio:
Hi @jvandersandt @vetheslav 👋 , if you are still experiencing this issue, we've made a change to this client's exception handling which should give you more context about the ApiResponseException you are getting.

As is the case with the issue linked by jvandersandt, this is often an issue with the curl configuration on your machine. Please update to v2.1.10 and inspect the context of the exception, as it should help you better debug the issue. Thanks.

ecoologic:
[...]

PR #322, opened 2017-04-11
Pass previous exception so that we don't lose it's message body

samgavinio:
/cc @zendesk/mintegrations

### Description

Underlying error message from curl are being lost because the original execption is not being passed along to the exception wrappers.

Previous exceptions that are returned as

```
PHP Fatal error: Uncaught exception 'GuzzleHttp\Exception\RequestException' with message 'Error completing request' in /var/www/sites/zendesk_api_client_php/vendor/guzzle/src/Exception/RequestException:71
```

should now look more informative. In the case of a curl SSL error, this will now look like:

```
PHP Fatal error: Uncaught exception 'GuzzleHttp\Exception\RequestException' with message 'cURL error 1: Protocol "https" not supported or disabled in libcurl (see http://curl.haxx.se.libcurl/c/libcurl-errors.html)' in /var/ww/sites/zendesk_api_client_php/vendor/guzzlehttp/guzzle/src/handler/CurlFactoryphp:187
```
This will allow people to figure out problems with their setup much easier, which in most cases is just a misconfiguration in cURL or missing php extensions:

* https://github.com/zendesk/zendesk_api_client_php/issues/316
* https://github.com/zendesk/zendesk_api_client_php/issues/248
* https://github.com/zendesk/zendesk_api_client_php/issues/289

### Risks
* [low] Could expose unwanted error messages.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -88,7 +88,7 @@ class Http
             }
             $response = $client->guzzle->send($request, $requestOptions);
         } catch (RequestException $e) {
-            $requestException = RequestException::create($e->getRequest(), $e->getResponse());
+            $requestException = RequestException::create($e->getRequest(), $e->getResponse(), $e);
             throw new ApiResponseException($requestException);
         } finally {
             $client->setDebug(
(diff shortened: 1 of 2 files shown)


samgavinio:
Struggled with getting this to go green on PHP5.5 due to a bug with re-throwing exception in try-finally blocks. Worked around it in the test.

Since folks, have signed off on this, merging this now.

PR #323, opened 2017-04-12
File upload stream

armen:
I needed the upload and multipart upload to support any PSR-7 stream.
This gives a lot of flexibility for instance now we can create a PSR-7 stream out of in-memory content and upload the content as an attachment.

I committed the modification independent of each-other so that you might cherry pick them if necessary but feel free to squash them.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -5,6 +5,7 @@ namespace Zendesk\API;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Psr7\LazyOpenStream;
 use GuzzleHttp\Psr7\Request;
+use Psr\Http\Message\StreamInterface;
 use Zendesk\API\Exceptions\ApiResponseException;
 use Zendesk\API\Exceptions\AuthException;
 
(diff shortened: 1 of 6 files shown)


tests/Zendesk/API/UnitTests/Core/MultipartUploadTest.php
```diff
@@ -0,0 +1,48 @@
...
+    public function setUp()
+    {
+        parent::setUp();
+        $this->dummyResource = new DummyResource($this->client);
+    }
+
+    /**
+     * Test multipart upload with streams
+     */
+    public function testUploadStream()
```
samgavinio: It appears this is already covered in here in the [resource test]( https://github.com/zendesk/zendesk_api_client_php/blob/9029142bdc33f47e40aac2654e9f4a214ec437d4/tests/Zendesk/API/UnitTests/Core/ResourceTest.php#L232-L250). We can do without this one test.

src/Zendesk/API/Http.php
```diff
@@ -102,10 +105,6 @@ public static function send(
             $request->getBody()->rewind();
         }
 
-        if (isset($file)) {
```
samgavinio: ❤️

armen:
Hey @samgavinio Moved the test into ResourceTest as suggested then amended the pull request.

Cheers

samgavinio:
@armen This has now made the master branch. I've added samples here as well: https://github.com/zendesk/zendesk_api_client_php/pull/326

Unfortunately, we're in a feature freeze until the end of April so we can't tag and release a new version until then.

Thanks.

armen:
Cool, thanks.

PR #324, opened 2017-04-20
[MI-1528] HelpCenter Articles Search

kcasas:
/cc @zendesk/mintegrations

### Description
* HelpCenter Articles Search endpoint made available

### TODO
* **DO NOT MERGE**
* Wait for April Feature freeze to end before merging and then reply to [support ticket](https://support.zendesk.com/agent/tickets/2531797)

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-1528

### Risks
* [low] endpoint returns a 404 status code if the user did not activate his helpcenter yet

### DIFF
--- a/src/Zendesk/API/Traits/Resource/Search.php
+++ b/src/Zendesk/API/Traits/Resource/Search.php
@@ -21,8 +21,8 @@ trait Search
                 $this->resourceName = $this->getResourceNameFromClass();
             }
 
-            $this->setRoute(__FUNCTION__, $this->resourceName . '/search.json');
-            $route = $this->getRoute(__FUNCTION__, $params);
+            $route = $this->resourceName . '/search.json';
+            $this->setRoute(__FUNCTION__, $route);
         }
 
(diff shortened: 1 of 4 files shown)


src/Zendesk/API/Traits/Resource/Search.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
joseconsador: Looks good, but are there any other endpoints that use search?

kcasas: screen shot 2017-04-21 at 10 59 08 am
Yep, those are `Organizations`, `Requests`, `Users`, and `PhoneNumbers`. We can use this trait to refactor those `Resources` but did not take my liberty on it cause it is outside the scope of this task. Also I suspect in the future they have to add searches on other `Resources` too so might as well create it as a `Trait` now :)

kcasas:
@joseconsador

PR #340, opened 2017-07-13
[MI-1724] retry ssl issues by default

kcasas:
/cc @zendesk/mintegrations

### Description
* add guzzle middleware `RetryHandler`
* made retrying of ssl issue a default behaviour

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-1724

### Risks
* [low] might slow down api especially on parts that will trigger retries

### DIFF
--- a/src/Zendesk/API/Middleware/RetryHandler.php
+++ b/src/Zendesk/API/Middleware/RetryHandler.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Zendesk\API\Middleware;
+
+use GuzzleHttp\Exception\ConnectException;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\RetryMiddleware;
+
+class RetryHandler
+{
+    /**
+     * @var array $timeoutCodes list of timeout status codes: Request Timeout, Authentication Timeout, Gateway Timeout
+     */
+    private $timeoutCodes = [408, 419, 504];
+
+    private $options = [
+        'max' => 2, // limit of retries
+        'interval' => 300, // base delay between retries, unit is in milliseconds
+        'max_interval' => 20000, // maximum delay value
+        'backoff_factor' => 1, // backoff factor
+        'exceptions' => [ConnectException::class], // Exceptions to retry without checking retry_if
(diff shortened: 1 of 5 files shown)


src/Zendesk/API/Middleware/RetryHandler.php
```diff
@@ -0,0 +1,105 @@
...
+use GuzzleHttp\RetryMiddleware;
+
+class RetryHandler
+{
+    /**
+     * @var array $timeoutCodes list of timeout status codes: Request Timeout, Authentication Timeout, Gateway Timeout
+     */
+    private $timeoutCodes = [408, 419, 504];
+
+    private $options = [
```
samgavinio: We probably want to document this. Also does this conflict with the retry configurations in kragle?

kcasas: I'll document it. Nope, they basically have the same settings except with its default values.

tests/Zendesk/API/UnitTests/Middleware/RetryHandlerTest.php
```diff
@@ -0,0 +1,246 @@
...
+     * @dataProvider retryLimitProvider
+     */
+    public function testRetryLimit($limit, $success)
+    {
+        $config = [
+            'max' => $limit
+        ];
+
+        do {
+            $responses[] = new ConnectException('', new Request('GET', ''));
```
kcasas: I think it's okay and it saves as another line of code.

samgavinio: Just leaving this here FWIW: https://stackoverflow.com/a/8246214/3138223.

src/Zendesk/API/Middleware/RetryHandler.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $this->options = array_merge($this->options, $config);
+    }
+
+    /**
+     * Returns the function that will decide whether to retry the request or not.
+     *
+     * @return callable
+     */
+    public function decider()
```
miogalang: This name isn't very descriptive of what the function does

maybe something like `shouldRetryRequest` makes more sense?

Issue #343, opened 2017-07-28
Possible bug in PUTting tags


arthens:
Hey there,

we are trying to append tags to a ticket, but can't get it to work. Looking at the code we suspect it might be a bug, but we aren't sure.

This is the code we are using:

```php
$client->tickets($ticketId)->tags()->update(null, [$tag1, $tag2]);
```

but nothing happens. 

Debugging the code we realised that the client is sending a PUT request to `/tickets/{$id}/tags` with data 

```php
[
    'tag' => [
        $tag1,
        $tag2,
    ], 
]
```

The [API docs](https://developer.zendesk.com/rest_api/docs/core/tags) say the request content should contain `tags`, not `tag`, but the client is sending `tas` instead.

The culprit seems to be `objectName` in [Update](https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Traits/Resource/Update.php#L39). This trait expects the including class to define `objectName`, but [Tags](https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Resources/Core/Tags.php#L14) doesn't explicitly do that, and relies on [ResourceAbstract](https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Resources/ResourceAbstract.php#L69), which singularise it by default.

We can't see any obvious way of changing this behaviour, but we can see that using reflections to "fix" `objectName` solves our problem:

```php
$tags = $client->tickets($ticketId)->tags();
$reflection = new \ReflectionProperty(get_class($tags), 'objectName');
$reflection->setAccessible(true);
$reflection->setValue($tags, 'tags'); // the magic happens here

$tags->update(null, [$tag1, $tag2]);
```

Is this a bug? Or are we misusing the client? (we are on 2.20 btw)

Thank you

jmramos02:
Hello! Will try to recreate this issue. will get back to you asap!

arthens:
@jmramos02 did you have any luck reproducing it? Let me know if you need anything else from us

joseconsador:
@arthens Thanks for this, we've confirmed the issue on our end. We were not able to reproduce the `tas` issue though. We'll have a PR out for the `objectName`.

joseconsador:
PR merged. Marking this as closed.

arthens:
Thank you!

PR #344, opened 2017-08-07
[MI-1783] Set Tags resource object name

joseconsador:
/cc @zendesk/mintegrations

### Description

Set Tags resource object name to prevent magically singularizing it.

### References
* JIRA: https://zendesk.atlassian.net/browse/MI-1783
* Issue: #343 

### Risks
* [Low] A 'tag' return value could be returned and may not be read by the client in the future, because objectName is plural.

### DIFF
--- a/tests/Zendesk/API/LiveTests/TicketsTest.php
+++ b/tests/Zendesk/API/LiveTests/TicketsTest.php
@@ -303,4 +303,27 @@ class TicketsTest extends BasicTest
         $this->client->groups()->delete($group->id);
         $this->client->tickets()->delete($ticket->id);
     }
+
+    /**
+     * Test if tags are updated on ticket updated.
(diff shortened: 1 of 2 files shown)


tests/Zendesk/API/LiveTests/TicketsTest.php
```diff
@@ -303,4 +303,27 @@ public function testAssignTicketToGroup()
...
+     * Test if tags are updated on ticket updated.
+     *
+     * @throws \Zendesk\API\Exceptions\MissingParametersException
+     */
+    public function testTagsAdded()
+    {
+        $faker = Factory::create();
+
+        $tags = array_map(function () use ($faker) {
+            return $faker->word;
```
kcasas: `$faker->words`?

PR #356, opened 2017-10-02
REPL tool for debugging

kcasas:
/cc @zendesk/mintegrations

### Description
* Created a wrapper for psysh. Run the command `bin/console <subdomain> <email> <api token>`. This would create an instance of `Zendesk\Api\HttpClient` on $client variable.

### References
* psysh: http://psysh.org

### Risks
* [low] tool only used api token for authentication. Support for other authentication should follow.

### DIFF
--- a/bin/ConsoleCommand.php
+++ b/bin/ConsoleCommand.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Zendesk;
+
+require 'DocMethodMatcher.php';
+require 'Shell.php';
+
+use Psy\Configuration;
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Zendesk\API\Exceptions\ApiResponseException;
+use Zendesk\API\HttpClient;
+
+class ConsoleCommand extends Command
+{
+    /**
+     * Configure the command options.
+     *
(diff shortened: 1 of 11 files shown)


miogalang:
@kcasas Please fix the commits. Squash them maybe? I see a commit "urg"

bin/ConsoleCommand.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Zendesk;
+
+require 'DocMethodMatcher.php';
```
samgavinio: Any way to add these to the load path automatically?

bin/Shell.php
```diff
@@ -0,0 +1,20 @@
...
+use Psy\TabCompletion\Matcher\ObjectMethodsMatcher;
+
+class Shell extends \Psy\Shell
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function getTabCompletionMatchers()
+    {
+        $tabCompletionMatchers = parent::getTabCompletionMatchers();
```
samgavinio: Please document here a bit why there is a need to filter out ObjectMethodsMatcher and ObjectAttributesMatcher.

bin/ConsoleCommand.php
samgavinio: Can we move these source files within  the`src/Zendesk/API` namespace and just keep `console` in the bin directory?

samgavinio:
Super cool addition to the client. Some annotations in the matcher would be nice since it's sort of hard to follow. And do we need tests for this?

PR #358, opened 2017-10-09
Removes confusion of appending /token to usernames

zanderwar:
The Zendesk API docs state that if using a token, you should append /token to the username.

I myself personally appended /token to the username as there was no clarity without looking through the code that /token was appended automatically

This PR allows the user to both omit or include /token appendage


jmramos02:
can you add a test for this? :)

zanderwar:
That's a negative, I don't have time to wrap my head around the way the tests are setup. Feel free to add it to this PR yourself by pushing to the `patch-1` branch of [zanderwar/zendesk_api_client_php](https://github.com/zanderwar/zendesk_api_client_php)

I've invited you as a collaborator ;P

PR #365, opened 2017-11-27
Add CURLOPT_PROTOCOLS = CURLPROTO_HTTP | CURLPROTO_HTTPS

oligriffiths:
Without this option set, this is a security vulnerability where the remote server can respond with a non-http protocol such as FTP for example, and curl will happily just follow the redirect.
This needs applying to the V2 track also.


oligriffiths:
Thanks for merging this.
Can we get a new tag for this?
Thanks

PR #294, opened 2016-11-30
Create createOrganization

Lobaster:
An example of organization creation including tags and custom organization fields.

PR #366, opened 2017-12-05
add Organization creation example

Lobaster:
cleaning up my previous (poor) attempt at submitting this example:
https://github.com/zendesk/zendesk_api_client_php/pull/294

I believe I've addressed your style comments, and while working with SMichaels he recommended creating a fresh PR which I'm doing here. (closing 294)

### DIFF
--- a/samples/organizations/createOrganization
+++ b/samples/organizations/createOrganization
@@ -1,0 +1,28 @@
+<?php
+include("../../vendor/autoload.php");
+use Zendesk\API\HttpClient as ZendeskAPI;
+/**
+ * Replace the following with your own.
+ */
+$subdomain = "subdomain";
+$username  = "email@example.com";
+$token     = "6wiIBWbGkBMo1mRDMuVwkw1EPsNkeUj95PIz2akv";
+$client = new ZendeskAPI($subdomain);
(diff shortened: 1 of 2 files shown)


HARVS1789UK:
Unless I am missing something, there is no functionality to go with this example? I cannot see any references to 'create', 'create many' or 'create or update' for [organisations](zendesk_api_client_php/src/Zendesk/API/Resources/Core/Organizations.php) in this API client.

These would correspond to the [Zendesk API documentation](https://developer.zendesk.com/rest_api/docs/core/organizations#create-organization) should they exist.

Am I missing something and is this functionality likely to be added to the client API anytime soon?

Lobaster:
This example demonstrates syntax to create an organization within this API client, which it does currently support. Give it a try!
- The organization resource loads the default traits here: https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Resources/Core/Organizations.php#L7
- The default resource traits include create here: https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Traits/Resource/Defaults.php#L15
- and create lives here: https://github.com/zendesk/zendesk_api_client_php/blob/master/src/Zendesk/API/Traits/Resource/Create.php

PR #368, opened 2017-12-15
Add support of Ticket Field Options API in SDK

cmourizard:
Add support of Ticket Field Options API in SDK (https://developer.zendesk.com/rest_api/docs/core/ticket_fields#list-ticket-field-options)

### DIFF
--- a/src/Zendesk/API/Resources/Core/TicketFieldsOptions.php
+++ b/src/Zendesk/API/Resources/Core/TicketFieldsOptions.php
@@ -1,0 +1,163 @@
+<?php
+
+namespace Zendesk\API\Resources\Core;
+
+use Zendesk\API\Exceptions\CustomException;
+use Zendesk\API\Exceptions\MissingParametersException;
+use Zendesk\API\Http;
+use Zendesk\API\Resources\ResourceAbstract;
+use Zendesk\API\Traits\Resource\Defaults;
+
+/**
+ * The TicketFieldsOptions class exposes options methods for ticketsFields
+ */
+class TicketFieldsOptions extends ResourceAbstract
+{
+    use Defaults {
+        findAll as traitFindAll;
+        find as traitFind;
+        create as traitCreate;
+        update as traitUpdate;
+        delete as traitDelete;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $objectName = 'custom_field_option';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $objectNamePlural = 'custom_field_options';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUpRoutes()
+    {
+        $this->setRoutes(
+            [
+                'findAll'   => 'ticket_fields/{fieldId}/options.json',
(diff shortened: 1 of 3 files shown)


src/Zendesk/API/Resources/Core/TicketFieldsOptions.php
```diff
@@ -0,0 +1,163 @@
...
+
+        return $response;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function update($id = null, array $updateResourceFields = [])
+    {
+        print_r($id);
```
joseconsador: Please remove debug code.

jmramos02:
@cmourizard is there any reason you override find, create, update etc.?

cmourizard:
Hi @jmramos02 it's because the route is a specific subroute of endpoint with some specific argument by example:
- find is under an "id" route
- create/update need a specific argument named custom_field_option

src/Zendesk/API/Resources/Core/TicketFieldsOptions.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(array $params = [])
+    {
+        $this->addTicketFieldIdToRouteParams($params);
+
+        $response = Http::send(
```
samgavinio: Looks like doing `return $this->traitCreate($params);` should be sufficient here? From the [docs](https://developer.zendesk.com/rest_api/docs/core/ticket_fields#using-curl-create), the `create` method looks similar to other resources. `update` on the other hand, is indeed different.

src/Zendesk/API/Resources/Core/TicketFieldsOptions.php
```diff
@@ -0,0 +1,160 @@
...
+
+        if (empty($id)) {
+            $id = $this->getChainedParameter(get_class($this));
+        }
+
+        $updateResourceFields['id'] = $id;
+
+        $this->addTicketFieldIdToRouteParams($updateResourceFields);
+
+        $response = Http::send(
```
samgavinio: Using `$this->client->post` method in client looks like it can be used directly here. It's a only small abstraction though but might as well.

src/Zendesk/API/Resources/Core/TicketFieldsOptions.php
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * The TicketFieldsOptions class exposes options methods for ticketsFields
+ */
+class TicketFieldsOptions extends ResourceAbstract
+{
+    use Defaults {
+        findAll as traitFindAll;
+        find as traitFind;
+        create as traitCreate;
+        update as traitUpdate;
```
samgavinio: Can probably be removed if the default `update` method will not be used?

cmourizard:
Hello @joseconsador and @miogalang all modifications has been performed I let you review them.

cmourizard:
Sounds great @samgavinio thanks! 

Do you have an estimated date of merge @joseconsador?

samgavinio:
@cmourizard released as v2.2.4. Thanks!

PR #372, opened 2018-01-24
Implement HelpCenter Article creation

abackstrom:
This patch adds support for Article creation via chaining, e.g.:

    $article = $client->helpCenter->sections($sectionId)->articles()->create([
        'locale' => 'en-us',
        'title' => 'Smartest Fish in the World',
    ]);

The unit tests pass. I had some trouble running the "Live" tests against a Sandbox instance.

My team is relatively new to this code base, so I apologize in advance if any of this is off! Creation via chaining *seemed* to be the expected implementation.

### DIFF
--- a/samples/helpcenter/createArticles.php
+++ b/samples/helpcenter/createArticles.php
@@ -1,0 +1,31 @@
+<?php
+
+include('../../vendor/autoload.php');
+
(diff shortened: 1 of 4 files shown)


dpawluk:
Hi team, can we please review this PR and let @abackstrom know if it looks good or there are any other steps before we can approve?

PR #391, opened 2018-06-12
Expose rate limit and rate limit remaining

richardappamondo:
Expose rate limit and rate limit remaining when returning the data from Zendesk API calls.


src/Zendesk/API/Http.php
```diff
@@ -106,7 +106,21 @@ public static function send(
         }
 
         $client->setSideload(null);
+        
+        $body = json_decode($response->getBody());
 
-        return json_decode($response->getBody());
+        if ($body != null) {
```
joseconsador: I don't think we should be adding this header to the body. What if we pass the body as an object that also exposes the $response object's method? That way we don't break existing functionality.

richardappamondo: Yes, that makes sense, Currently did it this way as this was the quickest and easiest way to do it at the time. I understand what you mean, and will look and rewrite this section when i get chance.

Thanks,

PR #414, opened 2019-05-07
Fixed PHPDoc of Attachments::upload()

lyrixx:
see https://github.com/zendesk/zendesk_api_client_php/blob/62e1f8488a78694fa88fe4c084a8c485a55f2824/src/Zendesk/API/Resources/Core/Attachments.php#L69-L71


damienalexandre:
The PHP DocBlock is also wrong here:

https://github.com/zendesk/zendesk_api_client_php/blob/2d1659aeb4aea1229a50bc4c567f27ad8fe91722/src/Zendesk/API/Resources/Core/Attachments.php#L34-L41

lyrixx:
@dpawluk I don't understand what you did

lyrixx:
I open https://github.com/zendesk/zendesk_api_client_php/pull/415 because something went wrong here!

PR #421, opened 2019-07-16
add tags resource for Resource\Core\Users to work with user tags

rieznyk:
to be able to do
`$client->users($userId)->tags()->create()`


imacrayon:
Needing this to be merged 🙏  

Thanks!

ecoologic:

PR #424, opened 2019-07-25
Allow authentication with password

### LINKED ISSUE #423: Lack of support for authorization with password
(It is really a continuation of #1 issue, but that one was closed so long ago that no one might have seen my comment and so I'm creating a new one)

As in the title, currently package does not support authentication by password.

The package supports two authentication modes: `basic` and `oauth`. However, `basic` accepts only username/api token pair and append '/token' to username before creation Authorization header. 

I propose allowing as the second parameter to setAuth arrays containing also password instead of token so the following two would work:

```
$client->setAuth('basic', ['username' => $username, 'token' => $token]);
```
```
$client->setAuth('basic', ['username' => $username, 'password' => $password]);
```

Makinami:
Resolves #423

### DIFF
--- a/src/Zendesk/API/Utilities/Auth.php
+++ b/src/Zendesk/API/Utilities/Auth.php
@@ -60,8 +60,17 @@ class Auth
         $this->authStrategy = $strategy;
 
         if ($strategy == self::BASIC) {
-            if (! array_key_exists('username', $options) || ! array_key_exists('token', $options)) {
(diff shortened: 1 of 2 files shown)


ecoologic:
Hi,

 thank you for your request,
are you still interested in a solution for this issue?

ecoologic:
Closing due to inactivity. Re-open if needed.

PR #436, opened 2019-10-15
will update the PHPDoc block to add the @method tickets() to the User class

HVSoftware:
The method tickets was not exposed in the User class. For this reason PHPStorm and PHPStan had issues with this class.


ecoologic:
Thank you for your contribution ❤️

PR #440, opened 2019-11-19
Upgrade phpmd to latest version 

nogates:
### Description

Upgrade phpmd to latest version

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,7 +1,7 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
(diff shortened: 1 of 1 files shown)


sudo-plz:
Why use a lock file anyway for a library?
And if you need to use this version for phpmd, why not set in in composer.json?

nogates:
@sudo-plz We use the lock file to [ensure all developers use the same libraries versions](https://getcomposer.org/doc/02-libraries.md#lock-file) when working with this library. The `composer.lock` file is ignored when using this library inside other applications. 

And yes, I agree, we could have specifically require that version of the library. We may do that in the future. Thanks for the suggestion!

sudo-plz:
@nogates Thanks for the clarification!

PR #441, opened 2019-12-02
DynamicContentItemVariants: Fix objectName variables

### LINKED ISSUE #409: Cannot updateMany Dynamic Content Variants
**TL;DR**

I think that `$objectName` and `$objectNamePlural` constants in `DynamicContentItemVariants` class should be changed to `'variant'` and `'variants'` values respectively

**Issue**
If you try to use updateMany with dynamic content variants, you always get the following error from server:

`{"error":"UnknownAttributeError","description":"Invalid attribute: missing variants parameter"}`

After some debugging, I find out that this is what updateMany is sending to the server as body of the PUT request

```
{
    "items": [
        {
            "id": 23,
            "content": "Lorem Ipsum",
        }
    ]
}
```

While official documentation suggest a body formatted as follows

```
{
  "variants": [
        {
          "id": 23,
          "content": "Lorem Ipsum",
        },
    ]
}
```

As you can see (API/Traits/Resource/UpdateMany.php:35), updateMany uses class constant value as resource name, so I suppose that this is the issue

**How to Replicate**
Here my test code:

```
$client->dynamicContent()->items( $dc_id )->variants()->updateMany( [
    [
        'id' => $v_id,
        'content' => 'Lorm Ipsum'
    ]
] );
```

where `$dc_id` contains id of the Dynamic content field that I want to update and `$v_id` contains id of the specific variation that I need to change

**Possible Solution**
Changing constants values with suggested ones seems to work pretty fine for me

**Note**
I'm using version 2.2.7 of the library, but I can see that `DynamicContentItemVariants` class wasn't changed since this version

I would be happy to repeat my test with latest version if you think it could help

Zarthus:
Also adds the UPDATE method

Fixes #409

### DIFF
--- a/src/Zendesk/API/Resources/Core/DynamicContentItemVariants.php
+++ b/src/Zendesk/API/Resources/Core/DynamicContentItemVariants.php
@@ -8,6 +8,7 @@ use Zendesk\API\Traits\Resource\CreateMany;
 use Zendesk\API\Traits\Resource\Delete;
 use Zendesk\API\Traits\Resource\Find;
 use Zendesk\API\Traits\Resource\FindAll;
+use Zendesk\API\Traits\Resource\Update;
 use Zendesk\API\Traits\Resource\UpdateMany;
 
 /**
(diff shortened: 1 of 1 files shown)


Zarthus:
cc @nogates (not sure who other collaborators are)

VBeytok (approved):

ecoologic:

PR #467, opened 2021-05-18
Problem: Link to pagination doc is outdated

c-rack:
Solution: Replaced with new link


ecoologic:
Thank you for your contribution. Unfortunately, the link changed again. I will update with the correct one.

PR #472, opened 2021-07-27
Update Http.php

### LINKED ISSUE #470: Call to undefined function GuzzleHttp\Psr7\stream_for()
`Error: Call to undefined function GuzzleHttp\Psr7\stream_for()
#7 zendesk/zendesk_api_client_php/src/Zendesk/API/Http.php(69): Zendesk\API\Http::send
#6 zendesk/zendesk_api_client_php/src/Zendesk/API/HttpClient.php(494): Zendesk\API\HttpClient::post
#5 zendesk/zendesk_api_client_php/src/Zendesk/API/Resources/Core/Tickets.php(164): Zendesk\API\Resources\Core\Tickets::create`

RobinBoersma:
Deprecated: stream_for will be removed in guzzlehttp/psr7:2.0. Use Utils::streamFor instead.
Fixes #470

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -66,7 +66,7 @@ class Http
             $request                     = $request->withoutHeader('Content-Type');
             $requestOptions['multipart'] = $options['multipart'];
         } elseif (! empty($options['postFields'])) {
-            $request = $request->withBody(\GuzzleHttp\Psr7\stream_for(json_encode($options['postFields'])));
+            $request = $request->withBody(\GuzzleHttp\Psr7\Utils::streamFor(json_encode($options['postFields'])));
         } elseif (! empty($options['file'])) {
             if ($options['file'] instanceof StreamInterface) {
                 $request = $request->withBody($options['file']);


Gramzivi:
There is a problem with this commit because it won't work in Guzzle versions from 6.0 to 7.1 as it is written in composer.json

jmwebslave:
We too have run into this bug.  While I agree it won't work in Guzzle 6.0-7.1 the problem is anyone pulling in the code now won't have it work.  I'd suggest it's probably more critical to have it work for the current versions than it is for the old versions, since anyone running older versions won't have pulled this code anyway.

Gramzivi:
I think maybe the safest solution, for now, is just to update the composer.json file according to working versions. If you have an existing project, a new `composer install` would mess your system. And tag new version with this update, to enable others to make difference between the old and new version of guzzle. But the problem is that Zendesk does not maintain this project anymore, so I don't know if anyone can merge updates.

schonhoff:
I can't use Guzzle 1.x anymore because of other dependencies. Is there a solution without forking this package?

ecoologic:
This solution has already been implemented. Thank you for your contribution.

PR #469, opened 2021-07-14
change stream_for() to Utils::streamFor() to retain Guzzle 7.2 compatibility 

spencerrlongg:
`\GuzzleHttp\Psr7\stream_for` has been deprecated, `GuzzleHttp\Psr7\Utils::streamFor()` is the replacement method, and has been replaced below. 

https://docs.aws.amazon.com/aws-sdk-php/v3/api/function-GuzzleHttp.Psr7.stream_for.html

PR #476, opened 2021-10-29
Declare dependency on guzzlehttp/psr7

wouterj:
This package uses functions from `guzzlehttp/psr7`, which are removed in psr7 2.0. As no dependency was declared, it was possible to install psr7 2.0 together with this package, resulting in fatal errors. By declaring the dependency, we can limit psr7 to 1.x.

After #469 is merged, the version constraint can be updated to `^1.7 || ^2.0` (as the static method API used in that PR was introduced in 1.7.0).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,7 @@
   "require": {
     "php": ">=5.5.0",
     "guzzlehttp/guzzle": "^6.0 || ^7.0",
+    "guzzlehttp/psr7": "^1.7",
     "mmucklo/inflect": "0.3.*"
   },
   "require-dev": {


yoshdog (approved):

manted (approved):

schonhoff:
As mentioned in https://github.com/zendesk/zendesk_api_client_php/issues/470 maybe someone can update packagist? This would help a lot! :-)

GrahamCampbell:
Guzzle maintainer here. You should instead upgrade to replacement functions. They are available on 1.x and 2.x.

wouterj:
@GrahamCampbell yes, both actions should be done (declare dependency if you depend on something the package provides & use the BC layer for 2.0). That's what I meant with (#469 is doing the method replacement):

> After #469 is merged, the version constraint can be updated to ^1.7 || ^2.0 (as the static method API used in that PR was introduced in 1.7.0).

Seems like this was forgotten, I've submitted a new PR to fix the constraint: #480

PR #478, opened 2021-12-31
Fix PHP 8 introduced breaking change to call_user_func_array()

Afnisse:
`call_user_func_array()` is used by `HttpClient::getSideload(array $params = [])`,  and since PHP 8^  introduced named arguments, the use of an associative array as arguments, will break the code since the array_merge does not support named arguments.

The fix is changing this line:
https://github.com/zendesk/zendesk_api_client_php/blob/b451b743d9d6d81a9abf7cb86e70ec9c5332123e/src/Zendesk/API/HttpClient.php#L441

To this:
```
return call_user_func_array('array_merge', array_values($sideloads));
```

### DIFF
--- a/src/Zendesk/API/HttpClient.php
+++ b/src/Zendesk/API/HttpClient.php
@@ -438,7 +438,7 @@ class HttpClient
 
         if (! empty($sideloads = array_intersect_key($params, array_flip($sideloadKeys)))) {
             // Merge to a single array
-            return call_user_func_array('array_merge', $sideloads);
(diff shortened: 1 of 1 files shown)


yoshdog (approved):

manted (approved):

schonhoff:
As mentioned in https://github.com/zendesk/zendesk_api_client_php/issues/470 maybe someone can update packagist? This would help a lot! :-)

PR #507, opened 2023-09-01
Deprecate support

7ail:
We will no longer be supporting this package

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# THIS REPOSITORY IS NO LONGER SUPPORTED #
+
 # Zendesk PHP API Client Library #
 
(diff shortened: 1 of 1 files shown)


zFlabmonsta (approved):

BafS:
Why not? What is the rational behind it? What do you propose instead?

jabeler:
+1 can we get a bit more info on this decision, and what this means for Zendesk moving forward?

Obviously, everyone relying on this library is a paying Zendesk user.

Packagist shows this has 14M+ installs, which is not insignificant.
https://packagist.org/packages/zendesk/zendesk_api_client_php

PR #508, opened 2023-10-05
CI

ecoologic:
* Refresh language and libraries to use PHP 7.4
* Fixed some tests failing due to now using PHP 7.4
* Fixed some tests failing due to being broken in the first place
* Skipped some tests to get a green build
* Added CI running tests
* CI fails when it should, see [here](https://github.com/zendesk/zendesk_api_client_php/actions/runs/6426888352/job/17451723058?pr=508)
* Ignoring live specs for now

This is a decent milestone and a good time to split the work. Usually I don't leave `TODOs` in PRs but I will follow up with another PR to solve those issues.

```
OK, but incomplete, skipped, or risky tests!
Tests: 261, Assertions: 2284, Skipped: 6.
```

There's more than 6 tests to fix, the tough ones, but less than a dozen.

### DIFF
--- a/tests/Zendesk/API/UnitTests/Middleware/RetryHandlerTest.php
+++ b/tests/Zendesk/API/UnitTests/Middleware/RetryHandlerTest.php
@@ -205,6 +205,8 @@ class RetryHandlerTest extends BasicTest
      */
     public function testHttpClientRetry()
     {
+        $this->markTestSkipped('CBP TODO');
+
         $this->setUp();
         $config = $this->client->guzzle->getConfig();
         $sslException = new RequestException(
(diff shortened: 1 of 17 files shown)


tests/bootstrap.php
```diff
@@ -1,6 +1,9 @@
 <?php
 
+// TODO: Don't fail using deprecated code
+// error_reporting(E_ALL ^ E_DEPRECATED);
 error_reporting(- 1);
```
ecoologic, line 5: I will allow the use of deprecated code, at least as a milestone, but if I do that now the tests keep going and breaks the whole test execution, so I'll do that separately.

fbvilela (approved):
LGTM! 👍

tests/Zendesk/API/UnitTests/Middleware/RetryHandlerTest.php
```diff
@@ -36,7 +36,7 @@ public function testExceptionsRetry($exception, $success)
             }
         ];
         $client = $this->mockApiResponses([
-            new $exception('', new Request('GET', '')),
+            new $exception('', new Request('GET', ''), new Response()),
```
yoshdog, line 39: 👍

yoshdog (approved):

cryptomail:
🥳

PR #509, opened 2023-10-09
Red 1950 unskip tests

ecoologic:
Finishing what started in the #508. Unskipping the tests that work and fixing code that didn't. I also tested the code in a separate consumer.

```sh
Tests: 261, Assertions: 2297, Skipped: 3.
```

### DIFF
--- a/tests/Zendesk/API/UnitTests/Chat/AppsTest.php
+++ b/tests/Zendesk/API/UnitTests/Chat/AppsTest.php
@@ -12,19 +12,10 @@ class AppsTest extends BasicTest
      */
     public function testInstall()
     {
-        $this->markTestSkipped('CBP TODO assert');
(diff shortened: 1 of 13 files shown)


tests/Zendesk/API/UnitTests/Chat/AppsTest.php
```diff
@@ -28,7 +27,7 @@ public function testInstall()
             $this->client->chat->apps()->install($postFields);
         }, 'apps/installations.json', 'POST', [
             'postFields' => $postFields,
-            'apiBasePath' => '/api/chat/',
+            'apiBasePath' => '/api/v2/'
```
ecoologic, line 30: This is another case where I think the tests were failing even "back in the days". Very confusing.

Zendesk-NathanBellette (approved):
LGTM

PR #511, opened 2023-10-11
CBP tickets

ecoologic:
Introducing CBP Iterator for Tickets.

This will be generalized in coming PR.

Usage: See README and tests, but basically:

```php
$ticketsIterator = $client->tickets()->iterator();
foreach ($ticketsIterator as $ticket) {
    process($ticket) // Your implementation
}
```

Note that this PR doesn't address sorting and other params yet. It will be added in a different iteration.

### DIFF
--- a/src/Zendesk/API/Traits/Utility/TicketsIterator.php
+++ b/src/Zendesk/API/Traits/Utility/TicketsIterator.php
@@ -6,20 +6,24 @@ use Iterator;
 
 class TicketsIterator implements Iterator
 {
-    const DEFAULT_PAGE_SIZE = 100;
     private $client;
     private $position = 0;
     private $tickets = [];
-    private $nextCursor = null;
+    private $afterCursor = null;
+    private $pageSize;
+    private $started = false;
 
-    public function __construct($client)
+    public function __construct($client, $pageSize = 2)
     {
         $this->client = $client;
-        $this->loadTickets();
+        $this->pageSize = $pageSize;
     }
 
     public function current()
     {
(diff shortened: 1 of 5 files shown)


src/Zendesk/API/Traits/Utility/TicketsIterator.php
```diff
@@ -0,0 +1,119 @@
...
+        $this->position = 0;
+    }
+
+    /**
+     * @return bool True there is a current element after calls to `rewind` or `next`, possibly fetching a new page.
+     */
+    public function valid()
+    {
+        if (!isset($this->tickets[$this->position]) && (!$this->started || $this->afterCursor)) {
+            $this->getPage();
```
ecoologic, line 100: note we only fetch the next page if we're done with the previous one. we don't want to fetch all pages and then do all the processing. This is also in the readme.

src/Zendesk/API/Traits/Utility/TicketsIterator.php
```diff
@@ -0,0 +1,119 @@
...
+        $this->resources = $resources;
+        $this->pageSize = $pageSize;
+    }
+
+    /**
+     * @return Ticket The current ticket, possibly fetching a new page.
+     */
+    public function current()
+    {
+        if (!isset($this->tickets[$this->position]) && (!$this->started || $this->afterCursor)) {
```
ecoologic, line 64: I just noted this can be abstracted. I'll do that in my next PR.

fbvilela (approved):

PR #512, opened 2023-10-18
CBP iterator for Users

ecoologic:
Now using CBP for Tickets and Users. TicketIterator becoming CbpIterator.

PS: There's a major refactoring coming to accommodate OBP endpoints, this is a minor stepping stone. Most of the changes are renaming and extractions.

### DIFF
--- a/src/Zendesk/API/Traits/Utility/CbpIterator.php
+++ b/src/Zendesk/API/Traits/Utility/CbpIterator.php
@@ -14,8 +14,13 @@ class CbpIterator implements Iterator
      */
     public const DEFAULT_PAGE_SIZE = 100;
 
+
     /**
-     * @var \stdClass implementing the iterator ($this), with findAll() defined
+     * @var string The root key in the response with the page resources (eg: users, tickets).
+     */
+    private $resourcesKey;
+    /**
+     * @var \stdClass implementing the iterator ($this), with findAll() defined.
      */
     private $resourcesRoot;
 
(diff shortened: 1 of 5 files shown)


src/Zendesk/API/Traits/Utility/CbpIterator.php
```diff
@@ -0,0 +1,133 @@
...
+     */
+    private function getPage()
+    {
+        $this->started = true;
+        $params = ['page[size]' => $this->pageSize];
+        if ($this->afterCursor) {
+            $params['page[after]'] = $this->afterCursor;
+        }
+        $response = $this->resourcesRoot->findAll($params);
+        $this->page = array_merge($this->page, $response->{$this->resourcesKey});
```
ecoologic, line 122: now using `resourcesKey` (eg: `tickets` or `users`) we can get the data in the response from the right place.

src/Zendesk/API/Traits/Utility/CbpIterator.php
```diff
@@ -0,0 +1,132 @@
...
+     * @var \stdClass implementing the iterator ($this), with findAll() defined.
+     */
+    private $resourcesRoot;
+
+    /**
+     * @var int The current position in the page.
+     */
+    private $position = 0;
+
+    // TODO: page is actually growing every call, should be only one page (reset position)
```
ecoologic, line 31: currently up for debate. I'll update if we decide, in my next PR.

PR #518, opened 2023-11-08
RED-2002 Iterator Associations

ecoologic:
* Allows lib consumers to define what method they want to call for their iterator
* Better docs on upgrading

### DIFF
--- a/CBP_UPGRADE_GUIDE.md
+++ b/CBP_UPGRADE_GUIDE.md
@@ -56,35 +56,7 @@ If this is your situation, **you will need to change the sorting order** to a su
 
 ## The new iterator
 
-The most efficient and elegant way to implement CBP is to use the newly provided iterator on your resources, for example:
-
-```php
-$params = ['my' => 'param1', 'extra' => 'param2'];
-$iterator = $client->tickets()->iterator($params);
-
-foreach ($iterator as $ticket) {
-    echo($ticket->id . " ");
-}
-```
-
-This will choose the right type of pagination and adapt your parameters for pagination and ordering to work with CBP.
-
-##### Iterator with params example
-
-```php
-$params = ['my' => 'param1', 'extra' => 'param2'];
-$iterator = $client->tickets()->iterator($params);
-
-foreach ($iterator as $ticket) {
-    echo($ticket->id . " ");
-}
-```
-
-* Change page size with: `$params = ['page[size]' => 5];`
-* Change sorting with: `$params = ['sort' => '-updated_at'];`
-  * Refer to the docs for details, including allowed sort fields
-* Combine everything: `$params = ['page[size]' => 2, 'sort' => 'updated_at', 'extra' => 'param'];`
-
+Please refer to the [README](./README.md#iterator-recommended).
 
 ## Parallel requests
 
--- a/README.md
+++ b/README.md
@@ -123,10 +123,6 @@ $tickets = $client->tickets()->sideload(['users', 'groups'])->findAll();
 
 ### Pagination
 
-See the [API reference for pagination](https://developer.zendesk.com/api-reference/introduction/pagination).
-
-There are two ways to do pagination in the Zendesk API, **CBP (Cursor Based Pagination)** and **OBP (Offset Based Pagination)**. The recommended and less limited way is to use CBP.
-
 #### Iterator (recommended)
 
 The use of the correct pagination is encapsulated using the iterator pattern, which allows you to retrieve all resources in all pages, without having to deal with pagination at all:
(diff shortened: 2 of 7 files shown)


CBP_UPGRADE_GUIDE.md
```diff
@@ -55,35 +56,7 @@ If this is your situation, **you will need to change the sorting order** to a su
...
-foreach ($iterator as $ticket) {
-    echo($ticket->id . " ");
-}
-```
-
-* Change page size with: `$params = ['page[size]' => 5];`
-* Change sorting with: `$params = ['sort' => '-updated_at'];`
-  * Refer to the docs for details, including allowed sort fields
-* Combine everything: `$params = ['page[size]' => 2, 'sort' => 'updated_at', 'extra' => 'param'];`
-
```
ecoologic, line 86: Removing dup.

src/Zendesk/API/Http.php
```diff
@@ -90,6 +90,7 @@ public static function send(
             if ($client->getAuth()) {
                 list ($request, $requestOptions) = $client->getAuth()->prepareRequest($request, $requestOptions);
             }
+            // echo "\nExternal API call: " . $request->getMethod() . " " . $request->getUri() . "\n";
```
ecoologic, line 93: I'm keeping these commented code (above too) for now.

README.md
```diff
@@ -123,13 +123,11 @@ $tickets = $client->tickets()->sideload(['users', 'groups'])->findAll();
...
 
-See the [API reference for pagination](https://developer.zendesk.com/api-reference/introduction/pagination).
-
-There are two ways to do pagination in the Zendesk API, **CBP (Cursor Based Pagination)** and **OBP (Offset Based Pagination)**. The recommended and less limited way is to use CBP.
+Methods like `findAll()` call the API without any pagination parameter. If an endpoint supports pagination, only the first page will be returned. To fetch all resources, you need to make multiple API calls.
 
 #### Iterator (recommended)
 
-The use of the correct pagination is encapsulated using the iterator pattern, which allows you to retrieve all resources in all pages, without having to deal with pagination at all:
+The use of the correct pagination is encapsulated using the iterator pattern, which allows you to retrieve all resources in all pages, making multiple API calls, without having to worry about pagination at all:
```
Zendesk-NathanBellette, line 130: I am wondering if we should add the word type. Like:

> The use of the correct type of pagination...

Or maybe even re-write the paragraph a little to read as:

> Using the iterator pattern means that you will not need to try to determine the correct pagination type to use for a given endpoint. This is because the logic to handle the pagination is encapsulated within the implementation of the iterator pattern allowing you to retrieve all resources in all pages, making multiple API calls, without having to worry about pagination at all:

README.md
```diff
@@ -157,7 +155,28 @@ foreach ($iterator as $ticket) {
...
+    "resources_key", // The root key with resources in the response, usually plural and in underscore
+    [], // Extra params for your call
+);
+$iterator = PaginationIterator($client->resources(), $strategy);
+foreach ($ticketsIterator as $ticket) {
+    // Use as normal
+}
+```
+
+Where `resources` is the collection, eg: `$client->tickets()`. This can be useful for filter endpoints like [active automations](https://developer.zendesk.com/api-reference/ticketing/business-rules/automations/#list-active-automations). However, in this common case where you only need to change the method from `findAll()` to `findActive()` there's a better shortcut:
```
Zendesk-NathanBellette, line 173: I'm finding this part to be slightly confusing. I'm assuming that this

> Where `resources` is the collection, eg: `$client->tickets()`.

Is referring to this code on line 167
```
$iterator = PaginationIterator($client->resources(), $strategy);
```
We might want to make it more clear that this is what we are referring to. Maybe we could move the code from line 167 to be above the code sample?

We could say something like:

In the code sample below, where we have `resources()` as a placeholder for a collection like `tickets` or `users`

So if you want to iterate through the `tickets` collection, you would use the following line of code to retrieve the iterators for that collection:
```
$iterator = PaginationIterator($client->tickets(), $strategy);
```

Zendesk-NathanBellette (commented):
Just a couple more comments

README.md
ecoologic: This is slightly inaccurate as the iterator _pattern_ is just about looping and per se is not pagination oriented. This design is composing iterator and strategy patterns. A detail that's not worth mentioning.

Zendesk-NathanBellette: Ok, that makes sense. I will mark this as resolved.

ecoologic: I find it verbose. Same concept in twice as many lines.

README.md
ecoologic: OK, I see your point. I will amend but trying to keep it short.

Zendesk-NathanBellette (approved):
Look good! Nice work!

PR #519, opened 2023-11-09
Iterator handling errors

ecoologic:
The implementation already works as expected, I've added a test for completion.


cryptomail (approved):
includes error handling....super cool.

PR #522, opened 2023-11-15
Expose organization createOrUpdate method

petrcloverdx:


cryptomail (changes requested):
Would it be possible to add some tests here?
We appreciate the contribution and secondly....SORRY for being so delayed in the response :)

Let's add some tests that affirm the functionality here and we'll be Good To Go
❤️

PR #525, opened 2023-12-22
Bump v3.0.1 updates user agent

ecoologic:
We did not update the version in the code since v2.2.9. This makes the user agent incorrect. 🤦 

Screenshot 2023-12-22 at 10 33 39 am


Zendesk-NathanBellette (approved):
LGTM

PR #528, opened 2024-02-26
Adding count functionality to search

haakony:


ecoologic (commented):
Thank you for your contribution! ❤️ 

We require all code to be unit-tested, would you please be able to amend your PR?

Here's an example of a [unit test](https://github.com/zendesk/zendesk_api_client_php/blob/master/tests/Zendesk/API/UnitTests/Core/SearchTest.php).

Thank you

PR #542, opened 2024-10-31
add endpoint property

veshant:
fixes php 8.2 deprecated dynamic property creation


token-cjg (approved):
Nice one, thank you @veshant for this.  [Fewer warning messages](https://php.watch/versions/8.2/dynamic-properties-deprecated#:~:text=In%20PHP%208.2%20and%20later,lifetime%20of%20the%20application%20execution.) are definitely better 👍

PR #544, opened 2024-12-09
Fix php 8.4 nullable parameters deprecations

MrJuliuss:
Hi, 

Here are some fixes for PHP 8.4 deprecations about Implicity Nullable Parameter
More informations : https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core.implicitly-nullable-parameter

Regards


token-cjg (approved):
Thank you for this contribution!

PR #566, opened 2026-07-21
Add support for Guzzle 8

GrahamCampbell:
Guzzle 8.0.0 was released on July 20th. Lots of people want to use it, and thus popular packages in the ecosystem like this one need to be updated to support it, so here I am, adding support. Note that the removal of Guzzle 6 support (for 7 & 8 only) is justified because your minimum PHP version is 8.2.0 and Guzzle 6 does not support PHP 8 at all.

### DIFF
--- a/src/Zendesk/API/Http.php
+++ b/src/Zendesk/API/Http.php
@@ -3,9 +3,11 @@
 namespace Zendesk\API;
 
 use GuzzleHttp\Exception\RequestException;
+use GuzzleHttp\Exception\ResponseException;
 use GuzzleHttp\Psr7\LazyOpenStream;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Psr7\Utils;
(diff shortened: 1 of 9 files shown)


src/Zendesk/API/Http.php
```diff
@@ -95,8 +97,17 @@ public static function send(
             }
             $response = $client->guzzle->send($request, $requestOptions);
         } catch (RequestException $e) {
-            $requestException = RequestException::create($e->getRequest(), $e->getResponse(), $e);
+            // Guzzle 8 exposes responses only on ResponseException; Guzzle 7
+            // has getResponse() on every RequestException.
```
bikalbasnet: I think these Guzzle 8 comments are helpful for context in the PR description, but they feel like unnecessary noise in the codebase itself. Let's remove them from the code. WDYT?

GrahamCampbell: Done.

